PR #44, opened 2013-04-25
Added popup windows handling support

irs:
Implementation of proposition in https://github.com/minkphp/Mink/issues/158


stof:
Closing this as there is no Mink API yet for this feature since years. And adding a driver-specific API is a no-go according to our policy (and does not bring much value compared to accessing the underlying webdriver API directly as shown in https://github.com/minkphp/Mink/issues/158#issuecomment-230479602).

PR #168, opened 2014-07-10
Added support for getting css

grom358:
Add support for getting an element's computed CSS property. https://github.com/Behat/Mink/issues/581


aik099:
This can't be merged for several reasons:

* Mink is there to emulate user activity on the page and user can't get CSS of an element and do something with it
* The CSS getting feature has no analogs in other drivers

PR #179, opened 2014-09-16
slight update to readme which makes example complete

jeremyquinton:


stof:
Closing because we will remove this example entirely in a few days in favor of the doc in https://github.com/minkphp/docs/blob/master/drivers/selenium2.rst

PR #196, opened 2015-01-07
New functions for Driver

Kuzmenkovs:


aik099:
No new functions then, @Kuzmenkovs ?

Regarding code:

1. don't mix different topics in PR (alert management and screenshots)
2. if screenshot functionality is supported by other drivers, then corresponding PR's should be made there
3. there is separate discussion about cross-driver alert support at Behat/Mink#158

PR #197, opened 2015-01-24
Switch the repository structure to PSR-4

stof:
If you agree with this change @aik099, other drivers will follow


aik099:
Yes, we can strip `Behat\Mink\Driver` folder structure created for follow PSR-0. I don't know if we should touch WUnit driver because it has 2 different namespaces in it (Behat & WUnit).

stof:
I agree for WUnitDriver.

aik099:
Related to https://github.com/minkphp/Mink/pull/639.

PR #199, opened 2015-02-04
Update the driver to use findElementXpaths for Mink 1.7

stof:
This updates the driver to benefit from minkphp/Mink#633


aik099:
OK.

PR #201, opened 2015-02-21
Fix the window name test for the chrome driver

stof:
While the Firefox driver uses lowercase letters, the Chrome driver is using uppercase letters in the window name (yeah, I started running the testsuite with chromium rather than using firefox because of the Firefox 35 issue).


aik099:
OK

PR #203, opened 2015-03-05
Fixed dropTo() drop behaviour

### LINKED ISSUE #20: Drag and Drop not generating HTML5 "drop" event
See:  http://code.google.com/p/selenium/issues/detail?id=3604

Since the Selenium2Driver already uses Syn.js, I was thinking we could add synthetic events for the drag & drop, and/or update to a newer version of syn.js (from http://github.com/jupiterjs/syn)

senaria:
At the moment !this->wdSession()->buttonup() is  called before the javascript is prepared and executed, what caused the dropping to fail.

Closes #20

### DIFF
--- a/src/Behat/Mink/Driver/Selenium2Driver.php
+++ b/src/Behat/Mink/Driver/Selenium2Driver.php
@@ -887,7 +887,6 @@ JS;
         $this->wdSession->moveto(array(
             'element' => $destination->getID()
         ));
-        $this->wdSession->buttonup();
 
         $script = <<<JS
 (function (element) {
@@ -900,6 +899,8 @@ JS;
 }({{ELEMENT}}));
 JS;
         $this->withSyn()->executeJsOnElement($destination, $script);
+        
+        $this->wdSession->buttonup();
     }
 
     /**


aik099:
Have you created PR based on latest `master` branch?

senaria:
Yes it should, i forked the master repo and applied the change on that.

aik099:
I wonder why then `We can’t automatically merge this pull request.` message is shown instead of Merge button.

I see you've sent PR from `master` branch, which essentially will prevent you from sending more then 1 PR at a time.

Could you please add a test, that will prove that we no longer have this issue?

stof:
this is weird, because it tells us that there are conflicts.

aik099:
@stof , see header at https://github.com/netdudes/MinkSelenium2Driver (fork). It says it's forked from `Behat/MinkSelenium2Driver`. So it's older version of repo that we've kept for tarballs only.

The fork should be from https://github.com/minkphp/MinkSelenium2Driver instead.

aik099:
In fact tests should go to Mink repo so that we can check that all drivers support that drop event properly.

senaria:
Ah ok,

ill have a look. My colleague did the fork. I'll correct it, sorry for 
the problems.

aik099:
I'm not sure if can swap out forks and keep same PR. In next PR you'll be creating create topic branch from `master` and send PR from that topic branch please.

stof:
you don't need to recreate the fork actually (it is not important for this case). However, you should add a remote for the Mink repo locally (instead of the Behat repo), fetch the latest upstream master, and then rebase your PR on top of the latest master branch

aik099:
@senaria , it appears, that PR was closed by GitHub automatically as result of your fork removal. Please send another PR from correct fork/branch.

PR #204, opened 2015-03-05
Created accessor for desiredCapabilities

### LINKED ISSUE #202: Accessor for desiredCapabilities
Is there a reason there's no accessor for desiredCapabilities? (ie. getDesiredCapabilities())?

Thanks!

cmacnaughton-al:
Closes #202

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -103,7 +103,7 @@ class Selenium2Driver extends CoreDriver
     public function setDesiredCapabilities($desiredCapabilities = null)
     {
         if ($this->started) {
-            throw new DriverException("Unable to set desiredCapabilities: the session has already started");
+            throw new DriverException("Unable to set desiredCapabilities, the session has already started");
         }
 
         if (null === $desiredCapabilities) {
@@ -149,7 +149,7 @@ class Selenium2Driver extends CoreDriver
     /**
      * Gets the desiredCapabilities
      *
-     * @return array $desiredCapabilties
+     * @return array $desiredCapabilities
      */
     public function getDesiredCapabilities() {
         return $this->desiredCapabilities;
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -151,7 +151,8 @@ class Selenium2Driver extends CoreDriver
      *
      * @return array $desiredCapabilities
      */
-    public function getDesiredCapabilities() {
+    public function getDesiredCapabilities()
+    {
         return $this->desiredCapabilities;
     }
 
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -97,9 +97,15 @@ class Selenium2Driver extends CoreDriver
      * See http://code.google.com/p/selenium/wiki/DesiredCapabilities
      *
      * @param array $desiredCapabilities an array of capabilities to pass on to the WebDriver server
+     *
+     * @throws DriverException
      */
     public function setDesiredCapabilities($desiredCapabilities = null)
     {
+        if ($this->started) {
+            throw new DriverException("Unable to set desiredCapabilities: the session has already started");
+        }
+
         if (null === $desiredCapabilities) {
             $desiredCapabilities = self::getDefaultCapabilities();
         }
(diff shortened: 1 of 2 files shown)


aik099:
Please also add tests to cover changed code.

cmacnaughton-al:
I'm new to this process, so I'm not sure why that failed.  If there's something I need to address please let me know, I'd be happy to help!

tests/Custom/DesiredCapabilitiesTest.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Behat\Mink\Tests\Driver\Custom;
+
+use Behat\Mink\Driver\Selenium2Driver;
+use Behat\Mink\Tests\Driver\TestCase;
+
+class desiredCapabilitieTest extends TestCase
```
aik099: Class name should match filename. Probably you don't have test auto-loading issue because you're on Windows.

tests/Custom/DesiredCapabilitiesTest.php
```diff
@@ -0,0 +1,36 @@
...
+{
+
+    public function testGetDesiredCapabilties()
+    {
+        $driver = $this->getSession()->getDriver();
+        $this->assertNotEmpty($driver->getDesiredCapabilities(), 'desiredCapabilities empty');
```
aik099: I'm not sure if we need to assert content of desired capabilities at all. I thought about asserting:

1. that we have some capabilities at the start
2. what is set we're getting back
3. if we set during active session, then we get proper exception (exception class & message)

cmacnaughton-al: Sounds good.  I do have a question about this: for those first 2 tests you mention, do you mean to call `setDesiredCapabilities()` with some set of capabilities, then call `getDesiredCapabilities()` to verify what we get back is what we set?  If so, can you tell me how I can call `setDesiredCapabilities()` within in the test case?  From what I can tell, I would have to call `$this->getSession()->getDriver()->setDesiredCapabilities()`, but in doing so starts the session and thus makes setting the desiredCapabilities impossible/moot.  Is there somewhere else I need to call setDesiredCapabilities() during setup?

Again, I'm no expert on this, so any guidance you can provide would be helpful!

aik099: > If so, can you tell me how I can call setDesiredCapabilities() within in the test case?

If `setDesiredCapabilities` is a protected/private method, then you can set them indirectly during driver construction.

cmacnaughton-al: Where would I instantiate the driver?  It appears that it's already instantiated once phpunit hits the DesiredCapabilitiesTest class, and that's done from \Behat\Mink\Driver\TestCase.

aik099: Nobody prevents you from creating test case that doesn't use session at all and manually instantiate driver class.

cmacnaughton-al:
Please let me know if you have any insight as to why scrutinizer failed.  I'm actually working on a Mac, and all the tests pass fine.  I did correct the filename/classname issue.

src/Selenium2Driver.php
```diff
@@ -141,6 +147,15 @@ public function setDesiredCapabilities($desiredCapabilities = null)
     }
 
     /**
+     * Gets the desiredCapabilities
+     *
+     * @return array $desiredCapabilities
+     */
+    public function getDesiredCapabilities() {
```
stof: CS issue here. The curly brace should be on its own line

tests/Custom/DesiredCapabilitiesTest.php
```diff
@@ -0,0 +1,50 @@
...
+            'browser'           => 'firefox',
+            'name'              => 'Selenium2 Mink Driver Test',
+            'deviceOrientation' => 'portrait',
+            'deviceType'        => 'tablet',
+            'selenium-version'  => '2.45.0'
+        );
+
+        $driver = new Selenium2Driver('firefox', $caps);
+        $this->assertNotEmpty($driver->getDesiredCapabilities(), 'desiredCapabilities empty');
+        $this->assertTrue(gettype($driver->getDesiredCapabilities()) == 'array');
```
stof: use ``assertInternalType('array', ...)`` instead

tests/Custom/DesiredCapabilitiesTest.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->assertEquals($driver->getDesiredCapabilities(), $caps);
```
stof: the expected value should be the first argument. Otherwise error messages are confusing on failure.

aik099:
Looks good. Merging, thanks @cmacnaughton-al .

PR #206, opened 2015-03-23
Update Selenium to the latest version

stof:


aik099:
OK

stof:
I updated my commit to use 2.46, and added a second commit to remove the test failing on recent versions of Selenium and not bringing much value IMO.

aik099:
Please do not remove that. Fact, that `CurlExec` exception is now thrown and breaks the tests is a bug introduced in `php-webdriver` library in https://github.com/instaclick/php-webdriver/pull/58 (was released as 1.4.2 version, which we might want to ignore in out `composer.json`) and the fix https://github.com/instaclick/php-webdriver/pull/67 is pending review.

LionsAd:
1.4.3 was released with the patch reverted.

LionsAd:
So this should be good to go in

PR #207, opened 2015-04-02
Fix comment for setTimeouts

perk11:
Timeouts are in milliseconds, rather than microseconds.
Milliseconds are defined here: https://github.com/instaclick/php-webdriver/blob/master/lib/WebDriver/Session.php#L320


aik099:
:+1:

stof:
thanks

PR #209, opened 2015-06-25
[WIP] Use Ctrl+A to to clear text inputs and textareas instead of pressing backspace and delete repeatedly

### LINKED ISSUE #198: setValue does not work correctly with focused fields
This took me a long time to track down, and it's somewhat difficult to explain, but here goes...

This is causing tests to fail on my project because I'm testing pages that have certain fields auto-focused depending on user actions.

OS: Mac OS X 10.10
Browser: Firefox 31.4.0esr, 33.1.1, 34.0.5
Selenium: 2.44

Unfortunately it's difficult to create a pull request to demonstrate this because this repo uses tests and fixtures defined in one of it's vendors folders, which is outside of version control. So I'm going to have to post details here instead.

I had a lot of difficulty creating a failing test that would illustrate this issue, because I could not get the field to focus during testing using evaluateScript() or even by adding a simple button with an onclick="blablabla.focus()", because for some reason that didn't work either! *pulls hair out*.

The only way I can get this test to fail is to add a function that is called by a button... bizarre. Anyway, onto the details:

Add the following snippet to the end of the body tag in `/vendor/behat/mink/driver-testsuite/web-fixtures/basic_form.html`:

```html
    <button onclick="doFocus()">Focus</button>

    <script>
      function doFocus() {
        document.getElementById('lastn').focus();
      }
    </script>
```

Then drop the following into the GeneralTest class in `/vendor/behat/mink/driver-testsuite/tests/Form/GeneralTest.php`:

```php
    public function testFillFocusedTextField()
    {
        $session = $this->getSession();

        $session->visit($this->pathTo('/basic_form.html'));

        $page = $this->getSession()->getPage();
        $page->pressButton('Focus');

        $webAssert = $this->getAssertSession();
        $field = $webAssert->fieldExists('last_name');
        $field->setValue('Chekote');

        $this->assertEquals('Chekote', $field->getValue());
    }
```

Then run the test with `phpunit --filter "testFillFocusedTextField"`

It shouldn't matter where the caret is when the field is focused, because the Selenium2Driver::setValue() method prepends a backspace and a delete keystroke for each existing character to the value before passing it to the postValue method. This *should* mean that the existing value will be deleted regardless of where the caret is in the field.

Unfortunately, it is failing with the caret being anywhere except the very end of the existing value. It appears that when the field is already focused, there are two possible scenarios:

1. If the caret is at the very beginning of the existing value, then only the first "delete" keystroke is processed, and the new value inserted. So in the test above, this results in "Lastname" changing to "Chekoteastname", instead of the expected "Chekote".
2. If the caret is anywhere within the existing value, then all of the "delete" keystrokes seem to be completely ignored. A few examples:

```
Caret positon:    Result:
L|astname          Chekoteastname
La|stname          Chekotestname
Las|tname          Chekotetname
Last|name          Chekotename
```

I can fix this on my machine with a call to $element->clear() before calling the postValue(), but I understand that this probably introduces issues with events (which is why I'm guessing it was removed in 23ec6b4d4aaba1a6de8affe097a95682b90022bf).

Any advice / help would be greatly appreciated.

perk11:
This fixed #198 for me

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -656,11 +656,17 @@ JS;
         $value = strval($value);
 
         if (in_array($elementName, array('input', 'textarea'))) {
-            $existingValueLength = strlen($element->attribute('value'));
+            $deleteSequence = Key::CONTROL . 'a' . KEY::NULL_KEY . Key::COMMAND . 'a' . KEY::NULL_KEY;
+            if (strlen($value) === 0) {
+                $value = $deleteSequence . Key::BACKSPACE;
+            } else {
+                $value = $deleteSequence . $value;
+            }
             // Add the TAB key to ensure we unfocus the field as browsers are triggering the change event only
             // after leaving the field.
-            $value = str_repeat(Key::BACKSPACE . Key::DELETE, $existingValueLength) . $value . Key::TAB;
+            $value .= Key::TAB;
         }
+
 
         $element->postValue(array('value' => array($value)));
     }


src/Selenium2Driver.php
```diff
@@ -656,12 +656,18 @@ public function setValue($xpath, $value)
         $value = strval($value);
 
         if (in_array($elementName, array('input', 'textarea'))) {
-            $existingValueLength = strlen($element->attribute('value'));
+            $deleteSequence = Key::CONTROL . 'a' . KEY::NULL_KEY;
```
aik099: On Mac the shortcut for doing this is Command+A. We should detect OS and use corresponding shortcut.

perk11: What do you think would be a good way to detect OS? I think of adding `isMac()` method that would execute a JS `navigator.appVersion.indexOf("Mac")!==-1;` and cache the result, does this sound good?

aik099: Not sure. Maybe Selenium itself has reliable OS detection methods, that doesn't involve checking user agent string, that can be faked by setting corresponding header.

perk11:
@aik099 I updated it to press both Ctrl+A and Command+A, should work on Mac now.

src/Selenium2Driver.php
```diff
@@ -656,12 +656,18 @@ public function setValue($xpath, $value)
         $value = strval($value);
 
         if (in_array($elementName, array('input', 'textarea'))) {
-            $existingValueLength = strlen($element->attribute('value'));
+            $deleteSequence = Key::CONTROL . 'a' . KEY::NULL_KEY . Key::COMMAND . 'a' . KEY::NULL_KEY;
```
aik099: Sending both shortcuts doesn't sound quite right. Maybe we'd better stay with OS detection as you suggested, but use remote capabilities for retrieval of user agent like here: http://mashtips.com/get-system-os-browser-version-java-selenium-webdriver/

perk11: @aik099, yes pressing both keys causes "a" to be typed in Chrome, I'll look into OS detection.

stof:
Closing this as #198 has been marked as fixed on the Selenium side

PR #210, opened 2015-06-26
Switch to the container-based infrastructure on Travis

stof:
This uses the PHP builtin webserver to expose web fixtures rather than installing Apache. This relies on the fact that PHP 5.6 is installed on Travis even on jobs running with a different version, and so can be used for fixtures on all jobs (and we don't care about the PHP version used to run fixtures as we were not using the Travis one before anyway).

If this works fine, I will apply it on other driver repositories needing a webserver to expose fixtures (i.e. all except BrowserKitDriver)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: php
 
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 php: [5.3, 5.4, 5.5, 5.6, hhvm]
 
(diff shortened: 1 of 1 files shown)


stof:
Works (for the remaining failure, see #206)

aik099:
OK.

LionsAd:
This can be re-tested now and should work fine. Clever idea to always use the 5.6 webserver.

PR #211, opened 2015-06-28
Add vendor binaries to composer

afurculita:
This lists the "run-phantomjs" and "run-selenium" commands as vendor binaries in composer.json in order to be passed along to a user who installs the package


aik099:
No need to do this, because they are only used on Travis CI.

stof:
These files are only meant to be used by Travis, so exposing it to everyone installing the package does not make sense IMO. I'm closing this.

PR #215, opened 2015-09-01
Support litteral text for select options

alexandresalome:
By now, you cannot pass a select option label as an option.

This PR makes it possible to specify an option's text as selector.

PS : I don't get the meaning of ``not(@value)``, so I removed it.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -1040,7 +1040,7 @@ XPATH;
     {
         $escapedValue = $this->xpathEscaper->escapeLiteral($value);
         // The value of an option is the normalized version of its text when it has no value attribute
-        $optionQuery = sprintf('.//option[@value = %s or (not(@value) and normalize-space(.) = %s)]', $escapedValue, $escapedValue);
+        $optionQuery = sprintf('.//option[@value = %s or normalize-space(.) = %s]', $escapedValue, $escapedValue);
         $option = $element->element('xpath', $optionQuery);
 
         if ($multiple || !$element->attribute('multiple')) {


stof:
this is adding a new feature (allowing to select options by their text rather than their value). This must be submitted to the main Mink repo first to add a test in the driver testsuite ensuring that the feature works consistently.

regarding the ``not(@value)``, this is because Mink currently relies on the option value only. And when the attribute is not set, the value of an option is its text. Given that we are making an XPath query, we need to account for this case (the DOM implements it internally, but we cannot access DOM properties in XPath query, only HTML attributes)

stof:
@alexandresalome any plan to continue the work on this ?

alexandresalome:
Maybe later, I'll reopen a PR at this time.

dawehner:
Is it just me or does mink support it already?

```
        '%tagTextMatch%' => 'contains(normalize-space(string(.)), %locator%)',
...
  ,'option' => <<<XPATH
.//option
[(./@value = %locator% or %tagTextMatch%)]
XPATH
```

aik099:
It seems that way. I can't find exact version, where this was changed though.

dawehner:
@aik099 Should we reopen this PR or just create a new one?

aik099:
@dawehner , why you need to reopen PR considering that you think it was already fixed?

dawehner:
@aik099 Its not fixed for the selenium driver as its using its own selector somehow

aik099:
In that case, @dawehner , you can of course open another PR and address all items needed from https://github.com/minkphp/MinkSelenium2Driver/pull/215#issuecomment-136636134 comment.

dawehner:
@aik099 Sure, I'll give it a try!

PR #217, opened 2015-09-22
Clear `input` and `textarea` on `setValue`

cbe:
I had similar issues as described in #198, this should fix it and delete every character in the input or textarea.

- Send `Delete`- and `Backspace`-Keystrokes for every Character
  in the Input or Textarea
- Previous behaviour which sent all deletions + given value in
  one String didn't work

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -657,9 +657,15 @@ JS;
 
         if (in_array($elementName, array('input', 'textarea'))) {
             $existingValueLength = strlen($element->attribute('value'));
+            $deleteAndBackspace = Key::DELETE . Key::BACKSPACE;
+
+            for($i = 0; $i < $existingValueLength; $i++) {
+                $element->postValue(array('value' => array($deleteAndBackspace)));
+            }
+
             // Add the TAB key to ensure we unfocus the field as browsers are triggering the change event only
             // after leaving the field.
-            $value = str_repeat(Key::BACKSPACE . Key::DELETE, $existingValueLength) . $value . Key::TAB;
+            $value = $value . Key::TAB;
         }
 
         $element->postValue(array('value' => array($value)));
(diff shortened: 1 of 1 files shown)


aik099:
Making several `postValue` calls (as you did in this PR) would result in several `change` JS events triggered for that input/textarea. That is what we're trying to avoid by combining input clearing and new value setting into same `postValue` call.

This would also result in corresponding test, that checks `change` event being triggered only once, failure.

aik099:
> Previous behaviour which sent all deletions + given value in one String didn't work

Can you please elaborate more on this one? The tests, that checks existing value removal on new value setting pass, so this works as expected (or tests are incomplete).

cbe:
> Can you please elaborate (version of Selenium, Browser, Mink, Driver) more on this one?

Of course:
OS: Mac 10.10.5
Selenium Version: 2.47.1
Browser: Firefox 40.0.3
Mink: 1.7.0
Selenium2-Driver: 1.3.0

I have a form where one Input (type text) is already pre-filled. In my test I am trying to replace the value with a custom one. The driver only seems to send one DELETE-Sequence (or selenium only interprets it as a single one), resulting in the original string being shortened by one from the start. The new value then gets placed beforehand.

Example:
Existing Value: `Username150`
New Value: `foo`
Result: `foosername150`

> This would also result in corresponding test, that checks change event being triggered only once, failure.

What's the reason for this expected behavior?

aik099:
Sounds like a bug in Selenium Server version. Have you tried problem happens also in Chrome? or maybe earlier Firefox version?

cbe:
> Sounds like a bug in Selenium Server version. Have you tried problem happens also in Chrome? or maybe earlier Firefox version?

Okay, thanks for the fast response. I'm heading over to [this selenium issue](https://github.com/SeleniumHQ/selenium/issues/340) and see if there's more to find out

PR #218, opened 2015-09-30
Join desired & default capabilities

andrerom:
To avoid having to redefine all defaults in config.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -101,8 +101,11 @@ class Selenium2Driver extends CoreDriver
     public function setDesiredCapabilities($desiredCapabilities = null)
     {
         if (null === $desiredCapabilities) {
-            $desiredCapabilities = self::getDefaultCapabilities();
+            $desiredCapabilities = array();
         }
+
+        // Join $desiredCapabilities with defaultCapabilities
+        $desiredCapabilities = $desiredCapabilities + self::getDefaultCapabilities();
 
         if (isset($desiredCapabilities['firefox'])) {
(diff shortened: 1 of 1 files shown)


src/Selenium2Driver.php
```diff
@@ -101,9 +101,12 @@ protected function setBrowserName($browserName = 'firefox')
     public function setDesiredCapabilities($desiredCapabilities = null)
     {
         if (null === $desiredCapabilities) {
-            $desiredCapabilities = self::getDefaultCapabilities();
+            $desiredCapabilities = array();
         }
 
+        // Join $desiredCapabilities with defaultCapabilities
+        $desiredCapabilities = $desiredCapabilities + self::getDefaultCapabilities();
```
aik099: Does `+` operator handle nested arrays properly?

stof: ``+`` does the same than ``array_replace`` (except it works on PHP 5.2 while ``array_replace`` is 5.3+). So it does not replace recursively. But the default capabilities don't include any nested one anyway.

@andrerom I prefer using ``array_replace`` though. It makes the code more readable.

andrerom: no, but as there is none in getDefaultCapabilities, that is not a problem.

andrerom:
@stof @aik099 updated

andrerom:
@stof @aik099 updated

aik099:
Build failure is unrelated. Otherwise good to merge.

aik099:
Merging. Thanks, @andrerom .

PR #221, opened 2015-10-20
Switch to main window on soft session reset

### LINKED ISSUE #219: Switch to main window after session reset
Add call to `->switchToWindow(null);` to the `reset()` method of the driver so that session reuse won't cause popup madness, where next test/scenario starts in popup opened by previous one.

See https://github.com/Behat/MinkExtension/issues/220

aik099:
Closes #219

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -357,6 +357,7 @@ class Selenium2Driver extends CoreDriver
     public function reset()
     {
         $this->wdSession->deleteAllCookies();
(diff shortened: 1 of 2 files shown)


stof:
Looks like we have an issue with ``;`` in cookies

aik099:
Of course we do. I've already told about it in PR that added these tests.

aik099:
And semicolon cookie test fails for Zombie as well.

PR #222, opened 2015-12-08
Update Selenium2Driver.php

pra-net:


aik099:
1. Before adding a feature it needs to be discussed.
2. The feature needs to be added for all drivers that can support it.

For this particular feature (that you've forgot to explain in PR description) there is an issue already: https://github.com/minkphp/Mink/issues/158 . Until consensus is reached in there no alert/confirm/prompt related changes will be made in the drivers.

You're welcome to participate in mentioned issue's discussion.

src/Selenium2Driver.php
```diff
@@ -325,7 +325,22 @@ private function applyTimeouts()
             throw new DriverException('Error setting timeout: ' . $e->getMessage(), 0, $e);
         }
     }
+      /**
+     * {@inheritdoc}
```
aik099: The DocBlock can't be inherited because there are no such method called `acceptAlert` in parent class.

stof:
I'm closing this PR as it cannot be accepted as is. Please participate in the discussion if you want to make this feature appear (and just access the webdriver object directly in the meantime if needed as it is exposed)

PR #225, opened 2016-01-26
Update the Travis setup

stof:
- add testing on PHP 7
- use Selenium 2.49.1
- use phantomjs 2.1.1 for the Phantom 2 builds


aik099:
OK.

PR #226, opened 2016-01-27
Add safeguard for not found elements in the driver-specific tests

stof:
This avoid fatal errors when the element is not found, allowing the other tests to run.
This fatal error happens currently in PhantomJS builds


aik099:
OK.

PR #227, opened 2016-01-27
Fix the handling of cookies with semicolon

stof:


aik099:
OK. I guess PhantomJS is so non-standard, that it's tests will never pass.

Issue #228, opened 2016-02-08
Add support to get selenium and browser details.


rajeshtaneja:
With recent releases of Selenium we have been facing few fails because of click button (https://github.com/SeleniumHQ/selenium/issues/1202).

Also, as they have recently released lot of versions in last few months it's hard to keep up with the changes and direction they take. It will be helpful if there is a way to notify the user about selenium version being used and if that is supported.

Can we please have getSeleniumVersion and getBrowserName implementation for the same.

aik099:
And you plan to put added method calls all around test suite?

rajeshtaneja:
I am planning to use it in beforeScenario hook for the first JS sesssion, to notify if the selenium version is greater then supported version. If failure found then report them, and use the last supported version.

aik099:
But message would be misleading in case of 99% cases, because the particular problem that you've reported in https://github.com/SeleniumHQ/selenium/issues/1202 won't be the cause of each error.

rajeshtaneja:
The reason I am thinking of checking the selenium version is to inform user about the selenium version not being supported if greater then what we have tested with before release. To ensure the failure found is genuine, the dev should test it with recommended selenium version.

We have a custom progress format which display following before the execution starts:
`
Moodle 3.1dev (Build: 20160129), pgsql, ead6c29bb58b194861b05379bff21b583106a13e
`
`
Server OS "Linux", Browser: "firefox"
`
`
You are not using recommended selenium version. Please refer https://docs.moodle.org/dev/Acceptance_testing/Browsers#Working_combinations_of_OS.2BBrowser.2Bselenium
`

Probably it will be nice to implement getStatus() rather then being specific about selenium and implement https://code.google.com/p/selenium/wiki/JsonWireProtocol#/status

Please let me know if that make sense to you and I will update the patch accordingly for getStatus() and getBrowser() in mink and here.


In addition to above it would be nice to have getWebDriver() as well.

aik099:
Not sure if there is `status` analog in other drivers.

rajeshtaneja:
Sorry for the late response on this @aik099 

Do you think adding getStatus() in this driver is an option ? If yes, then I will be happy to create a patch for the same and rename this issue to just add getStatus()

aik099:
The `getBrowserName` might be a good addition to `DriverInterface`, but the `getSeleniumVersion` I'm not so sure anymore.

@stof , what you think about this?

[...]

PR #229, opened 2016-02-08
#228 Add support to get Selenium version and browser name

rajeshtaneja:


aik099:
The driver is called by Mink. These methods won't do any good for Mink.

The only use case I can imagine is that:

* driver is created somewhere else
* inside the test you need to detect which browser and Selenium server versions are used and do different things

If I'm correct, then by doing so you're forcing yourself to use Selenium Mink driver in the future as well.

rajeshtaneja:
I am planning to use this in beforeScenario for the first JS session to check if selenium version is greater then the supported version. $context->getSession()->getDriver()->getSeleniumVersion();

This is only used for putting the information using custom formatter.

aik099:
The `getBrowserName` function however is interesting one. Maybe it makes sense to add to `DriverInterface`, because other non-headless drivers might implement it as well. For example Sahi.

PR #230, opened 2016-02-17
fix check/uncheck for radio button

guiyomh:
I could not check a radio option with mink.
The problem came from the Driver.
So I propose this patch

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -671,7 +671,7 @@ JS;
     public function check($xpath)
     {
         $element = $this->findElement($xpath);
-        $this->ensureInputType($element, $xpath, 'checkbox', 'check');
+        $this->ensureInputType($element, $xpath, array('checkbox','radio'), 'check');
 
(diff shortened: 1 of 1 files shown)


aik099:
According to documentation (see http://mink.behat.org/en/latest/guides/interacting-with-pages.html) you should be using `selectOption` method to check a radio button instead of `check` method, which is for checkboxes only. Also user can't uncheck the radio button once it's checked and there `uncheck` method doesn't work on radio buttons as well.

stof:
closing this as it is indeed an invlaid usage. Interacting with a radio group means selecting an option between the different input of the groups. You cannot check them individually (and you cannot uncheck them at all).

For your next contributions to drivers, the workflow when fixing a driver bug is to always start by adding a test in the Mink driver-testsuite reproducing it, to forbid regressions and to ensure that all drivers are consistent. Btw, adding a new feature at the driver level also start in the Mink repo (with the base driver change and the consistency tests)

PR #233, opened 2016-02-27
Update Syn - 0.0.3

ysramirez:


aik099:
Any specific reason why not to use latest version on https://github.com/bitovi/syn , which is 0.1.4?

ysramirez:
0.1.x  is not compatible

stof:
do they have any kind of changelog or upgrading doc to explain what the incompatibilities are ?

stof:
One of the biggest change is the renaming from ``Syn`` to ``syn`` for their global variable.

It would still be great to upgrade to a maintained version of the library, even though it requires a bit more work

aik099:
We can use `(syn || Syn)` to access it then. No big deal for using latest version.

stof:
@aik099 no need to use this IMO. If we upgrade to the new lib, we can simply use the new name.

aik099:
OK.

ysramirez:
Please merge, this version works better, the change of Syn to syn, In think is a big change (other version of Mink)

aik099:
> the change of Syn to syn I think is a big change (other version of Mink)

This can only be problematic to whoever extended the driver and have created new methods that make use of `->withSyn()` because most of these methods have `Syn.` hardcoded which now would be invalid.

And again we can just do `window.Syn = window.syn;` to solve that problem for custom Selenium2 driver based drivers.

aik099:
> Please merge, this version works better

Please explain in detail what means `works better`. It's most unlikely that Syn developers would release new versions is that works worse, than previous version.

ysramirez:
The current version (from 2010 more or less) is older than this version (30 Jun 2014) and you don't need to upgrate all the library maybe for MinkSelenium2Driver 1.3.1

Obviously when you upgrade from v0.0.3 to v0.1.5 or v0.2.0 you need to renaming from Syn to syn, maybe for MinkSelenium2Driver 1.4.0

aik099:
The Syn is just an internal dependency. Upgrading it doesn't bring any new features to the driver, so no feature release is necessary. Just a regular patch release.

ysramirez:
For example in some project (complex project with a javascript and css) with the old version when we use "keypress" and auto drag happens, and thinks like that. 

Please tell us if you want upgrade this old library from 2010 without control version like you can see in https://github.com/bitovi/syn and https://github.com/bitovi/syn/releases to the v0.0.3 from 2014 like we proposed.

We want to send more PRs, but if this is proccess is so burocratic we prefered use other tool or a fork. 

Thanks

aik099:
> We want to send more PRs, but if this is proccess is so burocratic we prefered use other tool or a fork.

It's simple really. Currently outdated version of Syn library is used. Thanks for informing us about that. You're proposing to upgrade to another outdated version of it. Why not upgrade to the latest version. If something needs to be changed in the driver code (e.g. replacing `Syn` to `syn`), then this can be done as part of the PR.

After looking back at previous posts I can't find any specific reason why you won't want to use 0.1.x (or even 0.2.x, that was released 5 days ago) and still want to stick to to 0.0.x releases.

ysramirez:
Jump to 0.2.x is the end of the road, the natural way is:

| Syn  Version                    | MinkSelenium2Driver             |
|---------------------------------|---------------------------------|
| 0.0.x                           | patch the current version |
| 0.2.x + other functionalities*   | new MickSelenium2Driver version |

*Other functionalities like typing

aik099:
Ah, I see now. You want to do a Syn upgrade via several PRs, where this one upgrades to latest 0.0.x and next PR upgrades to 0.2.x?

ysramirez:
Yes!

aik099:
@stof , this can be merged now. Tests pass as well.

ysramirez:
@stof ?

stof:
Ok, merging this (even though I don't understand the need to perform the upgrade in 2 steps, given that all the upgrade work would end up in the same step anyway)

PR #235, opened 2016-03-22
Use real textContent

thomaschaaf:
getText is giving me trouble in that it is not returning text for complex xpaths. By using the same function as getHtml this fixes a lot of problems for me.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -509,7 +509,8 @@ class Selenium2Driver extends CoreDriver
     public function getText($xpath)
     {
         $text = $this->executeJsOnXpath($xpath, 'return {{ELEMENT}}.textContent;');
-        $text = (string) str_replace(array("\r", "\r\n", "\n"), ' ', $text);
+        $text = (string) str_replace(array("\r", "\r\n", "\n"), ' ', trim($text));
+        $text = preg_replace('/\s+/', ' ', $text);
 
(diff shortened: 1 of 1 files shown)


aik099:
This changes breaks a lot of tests. Please adjust PR so that tests will pass.

thomaschaaf:
This should fix everything caused by my change.

stof:
I don't understand the reason to stop using the WebDriver API in favor of doing things manually.

thomaschaaf:
I was having problems with getText simply not returning the correct things like here:
screenshot 2016-03-22 15 52 07
screenshot 2016-03-22 15 54 30

stof:
And if there was a case where it was broken previously, please start by sending a PR to add more tests in the driver-testsuite in Mink, so that we ensure that this case works fine in all drivers. Your PR does not ensure that anything work fine for your use case, as there is nothing preventing regressions in the future (someone could send a PR doing the exact opposite change saying that usign WebDriver is better, the testsuite would be green, and I would likely accept it as using the WebDriver API whenever possible is indeed a better idea than writing things in JS)

stof:
well, your screenshot is not something helping us reproduce the issue (or even understanding it)

aik099:
@thomaschaaf , please provide:

* HTML of the page you're opening via Selenium
* PHP code you're using to find element and get it's text

PR #240, opened 2016-07-10
Select option by text

spolischook:
Select option by text if value not match even if value exists

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -1058,8 +1058,8 @@ XPATH;
     private function selectOptionOnElement(Element $element, $value, $multiple = false)
     {
         $escapedValue = $this->xpathEscaper->escapeLiteral($value);
-        // The value of an option is the normalized version of its text when it has no value attribute
-        $optionQuery = sprintf('.//option[@value = %s or (not(@value) and normalize-space(.) = %s)]', $escapedValue, $escapedValue);
+        // The value of an option is value attribute or the normalized version of its text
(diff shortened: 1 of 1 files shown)


spolischook:
This version will work with this kind of select:
```html
<select name="priority">
    <option value="low">Low</option>
    <option value="normal">Normal</option>
    <option value="high">High</option>
</select>
```
with:
```php
$selectElement->setValue('High');
```

spolischook:
Or another common example:
```html
<select name="interval_unit">
    <option value="M">minutes</option>
    <option value="H">hours</option>
    <option value="D">days</option>
    <option value="W">weeks</option>
</select>
```
with:
```php
$select->setValue('days');
```

spolischook:
Above examples works with new version of optionQuery. No BC broken

spolischook:
See test results for old optionQuery

tests/Custom/SelectTest.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Behat\Mink\Tests\Driver\Custom;
+
+use Behat\Mink\Tests\Driver\TestCase;
+
+class SelectTest extends TestCase
```
aik099: As I've mentioned in #241 the test should go into `/driver-testsuite/tests/Form/SelectTest.php` file and be submitted as Mink repo PR.

For example looking as Zombie driver code I can see, that in there setting value by option text is also not supported.

spolischook: ok

spolischook: Moved to Mink https://github.com/minkphp/Mink/pull/710

PR #241, opened 2016-07-10
[WIP] Test set value for select field

spolischook:


spolischook:
Fail test for https://github.com/minkphp/MinkSelenium2Driver/pull/240

tests/Custom/SelectTest.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Behat\Mink\Tests\Driver\Custom;
```
aik099: It should be part of form tests, because other similar stuff is there.

spolischook: @aik099 did you mean Behat\Mink\Tests\Driver\Form\SelectTest ? This test is more situated for  Selenium2Driver class, because in Mink it test set value by NodeElement. However the problem is more deeper, because we have two way of setting value, it's confusing.

aik099: There should be no Driver-specific test in this case, because `setValue` for select should behave the same way for all drivers and therefore test should go to the `Mink` main repo instead.

tests/Custom/SelectTest.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Behat\Mink\Tests\Driver\TestCase;
+
+class SelectTest extends TestCase
+{
+    public function testSetValueToSelectField()
```
aik099: If this is same test as in #240, then this isn't needed.

tests/Custom/SelectTest.php
spolischook: No problem for me, move this test to Mink repo, however I assume that you understand that other drivers would be affected.

aik099: That is the point. All drivers should work the same way. It's not a bug in Selenium2 driver, that it's `setValue` implementation doesn't support selecting option by it's text. It's how it works consistently for all drivers.

spolischook:
https://github.com/minkphp/Mink/pull/710

tests/Custom/SelectTest.php
spolischook: Moved to Mink https://github.com/minkphp/Mink/pull/710

Issue #243, opened 2016-09-02
Key::TAB / postValue vs. Chrome 53


lifeofguenter:
It seems that with recent Chrome (v53) `Key::TAB` as used in here: https://github.com/minkphp/MinkSelenium2Driver/blob/master/src/Selenium2Driver.php#L681 will actually cause the character `\t` to be added into the input-box, instead of "hitting the tab key".

I can not reproduce it on a normal workstation setup (e.g. real desktop), but it does not work on a headless chrome setup (Ubuntu 15.10 / selenium-server-standalone-2.53.1.jar):

```
$ google-chrome-stable --version
Google Chrome 53.0.2785.92
```

```
$ chromedriver --version
ChromeDriver 2.23.409687 (c46e862757edc04c06b1bd88724d15a5807b84d1)
```

Can be reproduced (headless-browser) with the following snippet:

```
<?php

ini_set('display_errors', true);
error_reporting(E_ALL);

require 'vendor/autoload.php';

$driver = new \Behat\Mink\Driver\Selenium2Driver('chrome');
$session = new \Behat\Mink\Session($driver);

$session->start();

$session->visit('http://localhost/admin/login');
sleep(2);
$session->getPage()->fillField('email', 'test@test.test');
$session->getPage()->fillField('password', 'abcdef');
$session->getPage()->pressButton('login');
```

selenium-server is started like this:

```
screen -S selenium -d -m sh -c "DBUS_SESSION_BUS_ADDRESS=/dev/null xvfb-run -s '-screen 0 1024x768x24' -a java -jar /opt/selenium-server/selenium-server-standalone.jar > /home/bamboo/selenium.stdout.log 2> /home/bamboo/selenium.stderr.log"
```

```
$ java -version
java version "1.8.0_77"
Java(TM) SE Runtime Environment (build 1.8.0_77-b03)
Java HotSpot(TM) 64-Bit Server VM (build 25.77-b03, mixed mode)
```

Swapping out the browser from chrome to firefox (version 46.0.1) works, as well as removing that `Key::TAB` from the code above (however not sure if it is the right fix, as obviously it should be working?).

aik099:
1. maybe this is bug in Selenium, that changes in Chrome (in way how TAB is handled) are not handled properly
2. the TAB is sent to trigger `onchange` JS event

lifeofguenter:
Is this maybe related to: https://github.com/instaclick/php-webdriver/issues/70 ?

[...]

PR #244, opened 2016-09-15
Fixed unwanted trailing tab in chrome 53

adorin:
Closes https://github.com/minkphp/MinkSelenium2Driver/issues/243

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -682,8 +682,7 @@ JS;
         }
 
         $element->postValue(array('value' => array($value)));
-        $script = 'Syn.trigger("change", {}, {{ELEMENT}})';
-        $this->withSyn()->executeJsOnXpath($xpath, $script);
+        $this->trigger($xpath, 'change');
     }
 
     /**
@@ -821,8 +820,7 @@ JS;
      */
     public function focus($xpath)
     {
-        $script = 'Syn.trigger("focus", {}, {{ELEMENT}})';
-        $this->withSyn()->executeJsOnXpath($xpath, $script);
+        $this->trigger($xpath, 'focus');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Selenium2Driver.php
```diff
@@ -676,12 +676,11 @@ public function setValue($xpath, $value)
...
         if (in_array($elementName, array('input', 'textarea'))) {
             $existingValueLength = strlen($element->attribute('value'));
-            // Add the TAB key to ensure we unfocus the field as browsers are triggering the change event only
-            // after leaving the field.
-            $value = str_repeat(Key::BACKSPACE . Key::DELETE, $existingValueLength) . $value . Key::TAB;
+            $value = str_repeat(Key::BACKSPACE . Key::DELETE, $existingValueLength) . $value;
         }
 
         $element->postValue(array('value' => array($value)));
+        $this->blur($xpath);
```
aik099: The `change` event should be fired instead of `blur` to make tests pass.

aik099:
Also don't send PR from `master` branch, because you'll be unable to send 2 PR in parallel to same project.

src/Selenium2Driver.php
adorin: Done

aik099:
One test still fails. However I'm unsure if this is related to made changes or just to Firefox version updated on Travis CI, that doesn't play nice with Selenium Driver version that is used.

aik099:
For me latest stable Selenium Driver (2.53.0) + latest stable Firefox (48.0.1) doesn't work at all. When running test suite on Chrome I'm getting same `&quot;` error.

aik099:
I've created #245 issue for that failing test because it's not related.

aik099:
Otherwise change looks good if it solves problem you're having in both Chrome and Firefox.

sumthief:
Faced with the same problem on my current project. This solution works for me in Chrome 53.0.2785.116 (64-bit).

aik099:
The build failure is due different problem described in #245.

Otherwise tests are passing and I think this one is ready to be merged.

aik099:
Thanks @adorin , merging.

PR #246, opened 2016-09-28
Update the testsuite to use the new mink/driver-testsuite

stof:
This updates the testsuite to use https://github.com/minkphp/driver-testsuite

TODOs:
- [x] change the composer requirements once https://github.com/minkphp/driver-testsuite/pull/1 is merged

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     },
 
     "require-dev": {
-        "symfony/phpunit-bridge": "~2.7"
+        "mink/driver-testsuite": "dev-master"
     },
 
     "autoload": {
--- a/tests/Selenium2Config.php
+++ b/tests/Selenium2Config.php
@@ -16,7 +16,7 @@ class Selenium2Config extends AbstractConfig
      */
     public function createDriver()
     {
-        $browser = $_SERVER['WEB_FIXTURES_BROWSER'];
+        $browser = getenv('WEB_FIXTURES_BROWSER') ?: 'firefox';
         $seleniumHost = $_SERVER['DRIVER_URL'];
 
         return new Selenium2Driver($browser, null, $seleniumHost);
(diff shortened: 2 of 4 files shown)


stof:
Another benefit of this approach is that it will become possible to test drivers against their lower deps, while still using the uptodate testsuite, as it is decoupled from Mink itself.

composer.json
```diff
@@ -19,14 +19,21 @@
...
+    ],
+
     "require": {
         "php":                       ">=5.3.1",
         "behat/mink":                "~1.7@dev",
         "instaclick/php-webdriver":  "~1.1"
     },
 
     "require-dev": {
-        "symfony/phpunit-bridge": "~2.7"
```
stof: I removed this here, because I handled it in ``mink/driver-testsuite`` (I want to force drivers to handle reporting of deprecations in their testsuite). Btw, the driver testsuite also allows using the Symfony 3 version, which is better.

tests/Selenium2Config.php
```diff
@@ -16,7 +16,7 @@ public static function getInstance()
      */
     public function createDriver()
     {
-        $browser = $_SERVER['WEB_FIXTURES_BROWSER'];
+        $browser = getenv('WEB_FIXTURES_BROWSER') ?: 'firefox';
```
aik099: Why you're getting `WEB_FIXTURES_BROWSER` via `getenv`, but `DRIVER_URL` via `$_SERVER`?

stof: well, using ``getenv`` made it much easier to provide a default value (as I cannot use ``??`` on PHP 5). This is the reason (and we are already using getenv elsewhere in this class anyway).

aik099: OK.

stof:
I'm merging this, as fixing the existing test failure is out of the scope of this PR. I will work on other drivers tomorrow.

stof:
Actually, I forgot a change (missed because Mink still contains the tests). I will send another PR.

Issue #245, opened 2016-09-16
The "Selenium2Driver::getContent" method is html-decoding it's result


aik099:
The `GeneralTest::testAdvancedForm` test started to fail recently:

```
1) Behat\Mink\Tests\Driver\Form\GeneralTest::testAdvancedForm
Failed asserting that '<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"><head>
    <title>Advanced form save</title>
    <meta content="text/html;charset=UTF-8" http-equiv="Content-Type" />
</head>
<body>
array (
  'agreement' = 'on',
  'email' = 'ever.zet@gmail.com',
  'first_name' = 'Foo "item"',
  'last_name' = 'Bar',
  'notes' = 'new notes',
  'select_number' = '30',
  'sex' = 'm',
  'submit' = 'Register',
)
some_file.txt
1 uploaded file
</body></html>' contains "array (
  'agreement' = 'on',
  'email' = 'ever.zet@gmail.com',
  'first_name' = 'Foo &quot;item&quot;',
  'last_name' = 'Bar',
  'notes' = 'new notes',
  'select_number' = '30',
  'sex' = 'm',
  'submit' = 'Register',
)
some_file.txt
1 uploaded file".
/home/travis/build/minkphp/MinkSelenium2Driver/vendor/behat/mink/driver-testsuite/tests/Form/GeneralTest.php:220
```

Here is why:

1. not long ago as XSS protection we started to html-escape output on `.../vendor/behat/mink/driver-testsuite/web-fixtures/advanced_form_post.php` page used in above mentioned test
2. for unknown reasons the `source` API call of JsonWireProtocol (see https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidsource) that we're using for `Selenium2Driver::getContent` method (see https://github.com/minkphp/MinkSelenium2Driver/blob/master/src/Selenium2Driver.php#L475) started to html-decode (e.g. convert `&quot;` into `"`) it's result
3. now this test fails

I'm not sure how to approach this, because we can't html-encode all page source.

bwowk:
I think this is related to some Firefox API changes. The WebDriver call to get the page source relies on the Browser, which may or may decode HTML entities.

[...]

PR #247, opened 2016-09-30
Fix the path to tests

stof:
They should reference the new driver-testsuite package, not the old testsuite available in Mink, as it will be removed.


aik099:
OK.

stof:
We should really investigate this escaping issue.

aik099:
> We should really investigate this escaping issue.

I've already filed and issue about this some time ago: #245

PR #248, opened 2016-10-04
Implement type() method to use Syn.type. Fixed issue with key{press|d…

zeyus:
…own|up} not working on non-input elements.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -863,6 +863,21 @@ JS;
     }
 
     /**
+     * Simulates keyboard typing with Syn.type().
+     * See: https://github.com/bitovi/syn/blob/master/src/key.js#L133
+     *
+     * @param string     $xpath
+     * @param string     $text  example: "Firstname\tLast\tmail@example.org\r"
+     *
+     */
+    public function type($xpath, $text)
+    {
+      $text = json_encode($text);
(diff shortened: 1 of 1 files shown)


src/Selenium2Driver.php
```diff
@@ -863,6 +863,21 @@ public function keyUp($xpath, $char, $modifier = null)
...
 
     /**
+     * Simulates keyboard typing with Syn.type().
+     * See: https://github.com/bitovi/syn/blob/master/src/key.js#L133
+     *
+     * @param string     $xpath
+     * @param string     $text  example: "Firstname\tLast\tmail@example.org\r"
+     *
+     */
+    public function type($xpath, $text)
```
aik099: This method can be useful, but needs to be added to:

1. driver test suite
2. driver interface
3. at least to 2 other drivers

src/Selenium2Driver.php
```diff
@@ -863,6 +863,21 @@ public function keyUp($xpath, $char, $modifier = null)
     }
 
     /**
+     * Simulates keyboard typing with Syn.type().
```
stof: Mink drivers are abstractions. So the phpdoc of methods should absolutely not describe them by talking about their internal implementation.

and indeed, as @aik099 said, this method won't be accepted if it is not properly added to the Mink API (btw, people are not expected to use the driver directly)

PR #251, opened 2016-10-26
Speed up travis ci build

adorin:
Testing CI time when caching vendor/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: false
 cache:
   directories:
     - $HOME/.composer/cache/files
+    - vendor
 
 php: [5.3, 5.4, 5.5, 5.6, 7.0, 7.1, hhvm]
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,6 +5,7 @@ sudo: false
 cache:
   directories:
     - $HOME/.composer/cache/files
+    - vendor
```
stof: no need to cache the vendor folder, as the composer cache is already persisted (and will be used when switching to ``prefer-dist``.
And avoiding the cache on the vendor folder avoids invaliding the cache on each build (due to the randomness in the autoloader file) and also avoids weird issues in case the cache gets corrupt (a corruption in the composer cache will force a redownload in composer, while a corruption in the vendor folder will keep using the package with a broken content)

stof:
Closing in favor of #260 and #261 which do this in a different way

Issue #187, opened 2014-10-30
Problems with file uploads on remote hosts


bretrzaun:
I can't get remote file uploads to work using Mink with the Selenium2 driver and SauceLabs.
After some research I found the sources below saying one needs to use the "fileDetector".
I am not sure if Mink does already support this. I could not find anything mentioned in the docs and source.

Sources:
- https://saucelabs.com/resources/selenium-file-upload
- http://sauceio.wordpress.com/2014/01/08/remote-file-uploads-with-selenium-php/

aik099:
@robocoder , ping.

tompedals:
I resolved this issue with the following code:

```
$localFile = '/path/to/my/fixture.pdf';
$tempZip = tempnam('', 'WebDriverZip');
$zip = new \ZipArchive();
$zip->open($tempZip, \ZipArchive::CREATE);
$zip->addFile($localFile, basename($localFile));
$zip->close();

$remotePath = $this->getSession()->getDriver()->getWebDriverSession()->file([
  'file' => base64_encode(file_get_contents($tempZip))
]);

$this->attachFileToField('file_input', $remotePath);

unlink($tempZip);
```

I'll try and tidy this up to make it more useful and submit a PR.

aik099:
So the solution is to send actual file in archive to get it uploaded? What file the web server, which selenium goes to will receive, the archive or original file?

tompedals:
@aik099 The `$localFile` is what is attached to the field. Selenium seems to want it sent as a zip file.

tompedals:
I found this solution in the facebook/php-webdriver project.

https://github.com/facebook/php-webdriver/blob/7cfa09f0a50b5e93087a9dc0eb669bc559eb9961/lib/Remote/RemoteWebElement.php#L329

aik099:
Interesting. Could you please send a PR for that? I guess we need a way to determine which kind of upload we want to be used automatically as well.

tompedals:
The way it seems to be done elsewhere is by checking whether the specified file exists on the local filesystem. If it does then it will upload the file to Selenium and use the remote path. If it doesn't exist we assume it is a file on the Selenium server already so just send the keys.

https://github.com/facebook/php-webdriver/blob/7cfa09f0a50b5e93087a9dc0eb669bc559eb9961/lib/Remote/RemoteWebElement.php#L302

[...]

PR #252, opened 2016-10-27
Issue #187: Add support for uploading files to remote Selenium instance

pfrenssen:
Here's a proposed fix for #187. This works both on local and remote instances. I tested using a local installation of Selenium and a Docker instance.

### DIFF
--- a/bin/run-selenium-remote.sh
+++ b/bin/run-selenium-remote.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env sh
+set -e
+
+echo '    Downloading selenium'
+docker pull selenium/standalone-chrome
+echo '    Running selenium'
+docker run -d -p 4444:4444 --network=host selenium/standalone-chrome
--- a/bin/run-selenium-remote.sh
+++ b/bin/run-selenium-remote.sh
@@ -2,6 +2,6 @@
 set -e
 
 echo '    Downloading selenium'
-docker pull selenium/standalone-chrome
+docker pull selenium/standalone-firefox:2.53.1
 echo '    Running selenium'
-docker run -d -p 4444:4444 --network=host selenium/standalone-chrome
+docker run -d -p 4444:4444 --network=host selenium/standalone-firefox:2.53.1
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -13,6 +13,7 @@ namespace Behat\Mink\Driver;
 use Behat\Mink\Exception\DriverException;
 use Behat\Mink\Selector\Xpath\Escaper;
 use WebDriver\Element;
+use WebDriver\Exception\InvalidRequest;
 use WebDriver\Exception\NoSuchElement;
 use WebDriver\Exception\UnknownError;
 use WebDriver\Exception;
@@ -797,9 +798,15 @@ JS;
         // Upload the file to Selenium and use the remote path. This will
         // ensure that Selenium always has access to the file, even if it runs
         // as a remote instance.
-        $path = $this->uploadFile($path);
+        try {
+          $remote_path = $this->uploadFile($path);
+        }
+        catch (InvalidRequest $e) {
+          // File could not be uploaded to remote instance. Use the local path.
+          $remote_path = $path;
+        }
 
-        $element->postValue(array('value' => array($path)));
+        $element->postValue(array('value' => array($remote_path)));
     }
 
     /**
@@ -1129,11 +1136,16 @@ JS;
     /**
      * Uploads a file to the Selenium instance.
      *
+     * Note that uploading files is not part of the official WebDriver
+     * specification, but it is supported by Selenium.
+     * @see https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L533
+     *
      * @param string $path     The path to the file to upload.
      *
      * @return string          The remote path.
      *
      * @throws DriverException When the file is not found.
+     * @throws InvalidRequest  When the driver does not support file uploads.
      */
     public function uploadFile($path) {
         if (!is_file($path)) {
@@ -1148,14 +1160,22 @@ JS;
         $archive->addFile($path, basename($path));
         $archive->close();
 
-        // Note that uploading files is not part of the official WebDriver
-        // specification, but it is supported by Selenium.
-        // @see https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L533
-        $remote_path = $this->getWebDriverSession()->file([
-          'file' => base64_encode(file_get_contents($temp_filename)),
-        ]);
+        try {
+          $remote_path = $this->getWebDriverSession()->file(array('file' => base64_encode(file_get_contents($temp_filename))));
+        }
+        catch (InvalidRequest $e) {
+          // Catch the error so we can still clean up the temporary archive.
+        }
 
         unlink($temp_filename);
+
+        // If the file upload failed, then probably Selenium was not used but
+        // another web driver such as PhantomJS.
+        // @todo Support other drivers when (if) they get remote file transfer
+        // capability.
+        if (empty($remote_path)) {
+          throw new InvalidRequest();
+        }
 
         return $remote_path;
     }
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -1146,6 +1146,7 @@ JS;
      *
      * @throws DriverException When PHP is compiled without zip support.
      * @throws UnknownError    When an unknown error occurred during file upload.
+     * @throws \Exception      When a known error occurred during file upload.
      */
     public function uploadFile($path)
     {
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -13,7 +13,6 @@ namespace Behat\Mink\Driver;
 use Behat\Mink\Exception\DriverException;
 use Behat\Mink\Selector\Xpath\Escaper;
 use WebDriver\Element;
-use WebDriver\Exception\InvalidRequest;
 use WebDriver\Exception\NoSuchElement;
 use WebDriver\Exception\UnknownError;
 use WebDriver\Exception;
@@ -1138,7 +1137,6 @@ JS;
      *
      * Note that uploading files is not part of the official WebDriver
      * specification, but it is supported by Selenium.
-     * @see https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L533
      *
      * @param string $path     The path to the file to upload.
      *
@@ -1147,6 +1145,8 @@ JS;
      * @throws DriverException When PHP is compiled without zip support, or the file doesn't exist.
      * @throws UnknownError    When an unknown error occurred during file upload.
      * @throws \Exception      When a known error occurred during file upload.
+     *
+     * @see https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L533
      */
     public function uploadFile($path)
     {
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -794,6 +794,11 @@ JS;
         $element = $this->findElement($xpath);
         $this->ensureInputType($element, $xpath, 'file', 'attach a file on');
 
+        // Upload the file to Selenium and use the remote path. This will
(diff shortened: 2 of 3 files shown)


pfrenssen:
This still needs some work. There's a PHP error, but I also see that it now fails on a local instance of PhantomJS when running in webdriver mode.

bin/run-selenium-remote.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+set -e
+
+echo '    Downloading selenium'
+docker pull selenium/standalone-firefox:2.53.1
```
aik099: This uses Docker on Travis CI? I wonder how slow that might me.

pfrenssen: It's quite comparable it seems, the build takes about 30 seconds longer. It looks like this time is needed to download the docker image.

I think this is the most cost effective way of testing a remote instance of Selenium. If we want to try a "real" remote then we probably have to go with a hosted service like Browserstack.

pfrenssen:
This is ready for a review. The tests are failing but this is due to the known problem https://github.com/minkphp/MinkSelenium2Driver/issues/245

src/Selenium2Driver.php
```diff
@@ -1120,4 +1132,52 @@ private function trigger($xpath, $event, $options = '{}')
...
+     *
+     * @param string $path     The path to the file to upload.
+     *
+     * @return string          The remote path.
+     *
+     * @throws DriverException When the file is not found.
+     * @throws InvalidRequest  When the driver does not support file uploads.
+     */
+    public function uploadFile($path) {
+        if (!is_file($path)) {
```
aik099: There wasn't such check before. Is it even needed?

src/Selenium2Driver.php
```diff
@@ -1120,4 +1132,52 @@ private function trigger($xpath, $event, $options = '{}')
...
+     * @see https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L533
+     *
+     * @param string $path     The path to the file to upload.
+     *
+     * @return string          The remote path.
+     *
+     * @throws DriverException When the file is not found.
+     * @throws InvalidRequest  When the driver does not support file uploads.
+     */
+    public function uploadFile($path) {
```
aik099: The opening `{` of function declaration should be placed on next line.

P.S.
I wonder why Scrutinizer CI haven't picked that up.

src/Selenium2Driver.php
```diff
@@ -1120,4 +1132,52 @@ private function trigger($xpath, $event, $options = '{}')
...
+        if (!is_file($path)) {
+          throw new DriverException('Could not upload file, the file: ' . $path . '. was not found.');
+        }
+
+        // Selenium only accepts uploads that are compressed as a Zip archive.
+        $temp_filename = tempnam('', 'WebDriverZip');
+
+        $archive = new \ZipArchive();
```
aik099: Is `ZipArchive` class present in any PHP installation. If it's possible for some OS to have limited PHP installation, then we'd better check that such class exists and throw useful exception instead of getting Fatal Error on missing `ZipArchive` class.

src/Selenium2Driver.php
```diff
@@ -1120,4 +1132,52 @@ private function trigger($xpath, $event, $options = '{}')
...
+        $archive->open($temp_filename, \ZipArchive::CREATE);
+        $archive->addFile($path, basename($path));
+        $archive->close();
+
+        try {
+          $remote_path = $this->getWebDriverSession()->file(array('file' => base64_encode(file_get_contents($temp_filename))));
```
aik099: Use `$this->wdSession` directly (as done in other methods of the driver) instead of calling `$this->getWebDriverSession()` method.

src/Selenium2Driver.php
```diff
@@ -1120,4 +1132,52 @@ private function trigger($xpath, $event, $options = '{}')
...
+          // Catch the error so we can still clean up the temporary archive.
+        }
+
+        unlink($temp_filename);
+
+        // If the file upload failed, then probably Selenium was not used but
+        // another web driver such as PhantomJS.
+        // @todo Support other drivers when (if) they get remote file transfer
+        // capability.
+        if (empty($remote_path)) {
```
aik099: Better to replace with this:

```php
if (isset($e)) {
    throw $e;
}
```

The above version:

* doesn't recreate exception based
* just reuses exception caught before

src/Selenium2Driver.php
```diff
@@ -1120,4 +1132,52 @@ private function trigger($xpath, $event, $options = '{}')
...
+
+        $archive = new \ZipArchive();
+        $archive->open($temp_filename, \ZipArchive::CREATE);
+        $archive->addFile($path, basename($path));
+        $archive->close();
+
+        try {
+          $remote_path = $this->getWebDriverSession()->file(array('file' => base64_encode(file_get_contents($temp_filename))));
+        }
+        catch (InvalidRequest $e) {
```
aik099: Is `InvalidRequest` the only possible error from file upload operation?

Maybe we should catch and rethrow all exceptions?

src/Selenium2Driver.php
pfrenssen: Not strictly needed :)

src/Selenium2Driver.php
pfrenssen: Good point!

aik099: Also it seems, that you always are using `ZipArchive` class now.

To preserve BC I think, that:

* in case if class is absent and we know it's remote file (not exists locally) then throw `DriverException`
* in case case if class is absent and it's local file just upload file the old way without even creating `ZipArchive`

pfrenssen:
Addressed the remarks. This is ready for another look :)

src/Selenium2Driver.php
aik099: Then when `ZipArchive` class isn't available we should execute old code immediately without throwing an exception and let driver fail as it fails currently.

pfrenssen: Yes that is what is going to happen now. If `ZipArchive` is not there a `DriverException` is thrown, and this is caught by `attachFile()` which will then continue with the original (local) path. It will then fail exactly as it does now later on when the `sendKeys` command is executed.

src/Selenium2Driver.php
```diff
@@ -1120,4 +1132,63 @@ private function trigger($xpath, $event, $options = '{}')
         $script = 'Syn.trigger("' . $event . '", ' . $options . ', {{ELEMENT}})';
         $this->withSyn()->executeJsOnXpath($xpath, $script);
     }
+
+    /**
+     * Uploads a file to the Selenium instance.
+     *
+     * Note that uploading files is not part of the official WebDriver
+     * specification, but it is supported by Selenium.
+     * @see https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L533
```
aik099: Please move `@see` tag after `@throws`. Or in any place, where other methods in this project has `@see` tag placed.

aik099:
Except moving of `@see` tag the code looks good to me.

src/Selenium2Driver.php
```diff
@@ -13,6 +13,7 @@
 use Behat\Mink\Exception\DriverException;
 use Behat\Mink\Selector\Xpath\Escaper;
 use WebDriver\Element;
+use WebDriver\Exception\InvalidRequest;
```
aik099: The Scrutinizer CI says that this class import isn't used anymore: https://scrutinizer-ci.com/g/minkphp/MinkSelenium2Driver/inspections/89c102e8-c656-4124-b660-6129c7ec2f16/patches?status=new

pfrenssen:
Thanks for the quick reviews! I have addressed the remarks.

aik099:
1. have you actually checked, that new tests will fail without new library code changes?
2. when test suite is running locally on developer machine all will work properly as before, but of course no remote upload tests?

pfrenssen:
I don't fully understand the question.

I need this fix because we are running our tests on [continuousphp](https://continuousphp.com/) which recently switched to remote Selenium instances and our uploads were failing.

I have tested on my dev machine with local and remote instances of Selenium. It's now also working on continuousphp which it wasn't before.

pfrenssen:
Oh you probably mean to run the tests on Travis using a remote Selenium instance, but without the fix.

I made a branch containing only the test, but without the fix. Here is the relevant test result: https://travis-ci.org/pfrenssen/MinkSelenium2Driver/jobs/171671291

You can see that in the response the text `1 uploaded file` is missing after `some_file.txt`. This proves that without the fix the file is not uploaded to the remote instance.

Here is the same test result with the fix: https://travis-ci.org/pfrenssen/MinkSelenium2Driver/jobs/171596712

In this case you can see that the text `1 uploaded file` is present in the response, so the file is correctly uploaded to the remote instance.

aik099:
That's exactly what I've meant. Good.

aik099:
Merging, thanks @pfrenssen .

frankcarey:
This was such a pain for so long, THANK YOU!

src/Selenium2Driver.php
```diff
@@ -794,7 +794,18 @@ public function attachFile($xpath, $path)
         $element = $this->findElement($xpath);
         $this->ensureInputType($element, $xpath, 'file', 'attach a file on');
 
-        $element->postValue(array('value' => array($path)));
+        // Upload the file to Selenium and use the remote path. This will
+        // ensure that Selenium always has access to the file, even if it runs
+        // as a remote instance.
+        try {
+          $remote_path = $this->uploadFile($path);
```
stof: please use camelCase variables names

src/Selenium2Driver.php
```diff
@@ -794,7 +794,18 @@ public function attachFile($xpath, $path)
...
+        }
+        catch (\Exception $e) {
```
stof: should be on the previous line

src/Selenium2Driver.php
```diff
@@ -1120,4 +1131,64 @@ private function trigger($xpath, $event, $options = '{}')
...
+     *
+     * @return string          The remote path.
+     *
+     * @throws DriverException When PHP is compiled without zip support, or the file doesn't exist.
+     * @throws UnknownError    When an unknown error occurred during file upload.
+     * @throws \Exception      When a known error occurred during file upload.
+     *
+     * @see https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L533
+     */
+    public function uploadFile($path)
```
stof: should be private

src/Selenium2Driver.php
```diff
@@ -1120,4 +1131,64 @@ private function trigger($xpath, $event, $options = '{}')
...
+
+        if (!class_exists('ZipArchive')) {
+          throw new DriverException('Could not compress file, PHP is compiled without zip support.');
+        }
+
+        // Selenium only accepts uploads that are compressed as a Zip archive.
+        $temp_filename = tempnam('', 'WebDriverZip');
+
+        $archive = new \ZipArchive();
+        $archive->open($temp_filename, \ZipArchive::CREATE);
```
stof: you need to check the return value here, as ZipArchive uses the return value to handle errors instead of throwing exceptions

aik099:
@stof , thanks for reviewing, but PR is already merged. All the fixes should go as separate PR I guess.

stof:
yeah, I received a notification about this PR because of the comment, and I reviewed it while noticing it was merged only at the end.
Doing a new PR for the fixes is indeed the way to go (if it is not done at the time I work on releasing, I will do it myself, as I don't want to release with the method exposed publicly and so covered by BC)

pfrenssen:
Thanks for the review. I will open a followup to address the remarks. Will link it here when I make the PR.

bwowk:
Is there any roadmap for the next release yet? This feature by itself would deserve a release.

asprega:
+1. Is there a planned release for this feature?

kvdnberg:
+1 for release! We spend days in agony trying to fix this problem while testing CircleCI 2.0 with docker images (which can't share volumes). When I finally found this yesterday I was thrilled. One note: I had to debug it to get it to work because my docker php didn't have zip support and because the exception is caught and the 'local' file is then used, it seemed like this didn't fix the problem. So if anyone else is checking this fix and it doesn't seem to work, check your zip extension ;)

pfrenssen:
I made a separate issue that suggests to make a new release, the comments here are probably not going to be seen since this is already closed: #269

PR #253, opened 2016-12-15
Fix review remarks. Followup for #252.

pfrenssen:
This fixes the remarks made by @stof on #252. This is split off in a new PR since #252 is already merged.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -798,14 +798,13 @@ JS;
         // ensure that Selenium always has access to the file, even if it runs
         // as a remote instance.
         try {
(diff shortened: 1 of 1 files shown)


src/Selenium2Driver.php
```diff
@@ -1162,9 +1161,18 @@ public function uploadFile($path)
         $temp_filename = tempnam('', 'WebDriverZip');
```
stof: please use camelCase here too

pfrenssen:
Thanks for the review. Fixed it, and also `$remote_path` -> `$remotePath`.

Issue #254, opened 2016-12-28
Selenium 3 support


andrerom:
Any plans for selenium 3 support?

At least from [minimal testing](https://travis-ci.org/ezsystems/ezplatform/jobs/187229788#L738) it seems certain parts have changed:
```
WebDriver\Exception\UnknownCommand: POST /session/1e53463e-002e-472d-9027-24f8f303b6e8/moveto did not match a known command
```

stof:
Well, according to previous community reports about it, Selenium 3 should work fine when using Chrome, but the new Firefox driver is not yet feature complete.

Running our CI against Selenium 3 is on my TODO list.

andrerom:
This is indeed Firefox, will check with chrome, feel free to close if test passes.

JoaoGFarias:
Since [Selenium 3 now supports Safari (version 10](https://webkit.org/blog/6900/webdriver-support-in-safari-10/), would be good to test against it as well.

stof:
@JoaoGFarias see #257 for discussions about the CI setup

tractorcow:
https://github.com/instaclick/php-webdriver doesn't seem to support selenium 3, so perhaps a shift to something like https://github.com/facebook/php-webdriver?

tractorcow:
For reference I have been testing with firefox and geckodriver, and have had some partial success.

NicoHaase:
https://github.com/instaclick/php-webdriver/issues/80 should give another hint why any support of Selenium 3.5 is now really neccessary

tractorcow:
Yes that is the one issue I had with that driver. Perhaps instaclick will update to support both 2.x and 3.x?

aik099:
@tawfekov , you can open at https://github.com/instaclick/php-webdriver/issues then.

robocoder:
At Instaclick, we're drifting towards stability (rather than bleeding edge), thus, the infrequent updates to our OSS repos and forks.

Short-term it may be quicker to submit a PR to to `instaclick/php-webdriver`. Long-term `minkphp\MinkSelenium2Driver` should consider switching to a webdriver binding that's more actively maintained given that changes to the JSON Wire Protocol are becoming more frequent and likely to break backwards compatibility.

tractorcow:
I've got a mostly-usable version of mink2selenium working with facebook/webdriver, although I'm not actively working on it at the moment (currently sticking with selenium 2).

[...]

Issue #239, opened 2016-06-20
Failing test suite, Selenium 2.53


aik099:
Configuration:

* OS: Yosemite 10.10.5
* Google Chrome: 51.0.2704.84 (64-bit)
* Selenium Server: 2.53

When running driver tests in there I have following failures:

```
There was 1 error:

1) Behat\Mink\Tests\Driver\Js\ChangeEventTest::testSetValueChangeEvent with data set "file" ('the-file', 'from empty', 'from existing')
WebDriver\Exception\UnknownError: unknown error: path is not absolute: from empty
  (Session info: chrome=51.0.2704.103)
  (Driver info: chromedriver=2.21.371459 (36d3d07f660ff2bc1bf28a75d1cdabed0983e7c4),platform=Mac OS X 10.10.5 x86_64) (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 20 milliseconds
Build info: version: '2.53.0', revision: '35ae25b', time: '2016-03-15 17:00:58'
System info: host: 'alex-mbp', ip: '192.168.1.6', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.10.5', java.version: '1.8.0_74'
Driver info: org.openqa.selenium.chrome.ChromeDriver
Capabilities [{applicationCacheEnabled=false, rotatable=false, mobileEmulationEnabled=false, chrome={chromedriverVersion=2.21.371459 (36d3d07f660ff2bc1bf28a75d1cdabed0983e7c4), userDataDir=/var/folders/zr/z961h7s124z_qq5s4c4_j4000000gn/T/.org.chromium.Chromium.nXHJDJ}, takesHeapSnapshot=true, databaseEnabled=false, handlesAlerts=true, hasTouchScreen=false, version=51.0.2704.103, platform=MAC, browserConnectionEnabled=false, nativeEvents=true, acceptSslCerts=true, locationContextEnabled=true, webStorageEnabled=true, browserName=chrome, takesScreenshot=true, javascriptEnabled=true, cssSelectorsEnabled=true}]
Session ID: 68f84e17342891af646e900968ac9a26

/path/to/repo-clone/vendor/instaclick/php-webdriver/lib/WebDriver/Exception.php:155
/path/to/repo-clone/vendor/instaclick/php-webdriver/lib/WebDriver/AbstractWebDriver.php:140
[...]

PR #255, opened 2017-01-03
Improve compatibility with GeckoDriver when using Selenium 3

stof:
moveto is not part of the W3C Webdriver spec, and so GeckoDriver does not implement it.
Our click() implementation uses it to ensure that the element is scrolled into the viewport before clicking on it, due to Selenium requirements.
Instead of forbidding the click entirely when using Geckodriver, this now attempts the click directly when moving is unsupported, hoping that the driver has not implemented the restriction when they don't have a way to force the move yet.

Refs #254

Other methods using ``moveto`` are not changed like this as they rely on the cursor position too (when using other non-standardized APIs).

Before:
Tests: 171, Assertions: 300, Errors: 46, Failures: 4, Skipped: 14.

After:
Tests: 172, Assertions: 375, Errors: 11, Failures: 6, Skipped: 13.

(and one of the failures and one of the errors are the one from #239)

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -14,6 +14,7 @@ use Behat\Mink\Exception\DriverException;
 use Behat\Mink\Selector\Xpath\Escaper;
 use WebDriver\Element;
 use WebDriver\Exception\NoSuchElement;
+use WebDriver\Exception\UnknownCommand;
 use WebDriver\Exception\UnknownError;
 use WebDriver\Exception;
 use WebDriver\Key;
@@ -764,7 +765,13 @@ JS;
 
     private function clickOnElement(Element $element)
     {
-        $this->wdSession->moveto(array('element' => $element->getID()));
+        try {
+            // Move the mouse to the element as Selenium does not allow clicking on an element which is outside the viewport
+            $this->wdSession->moveto(array('element' => $element->getID()));
+        } catch (UnknownCommand $e) {
+            // If the Webdriver implementation does not support moveto (which is not part of the W3C WebDriver spec), proceed to the click
+        }
+
         $element->click();
     }
 


aik099:
Maybe we should do JS scrolling if `moveto` call fails just to make test pass?

stof:
@aik099 failing tests are *not* related to the element being clicked out of view, but to other unimplemented APIs:

- usage of non-W3C webdriver methods for drag&drop, right click and double click (which all rely on moveto btw)
- usage of moveto for ``mouseOver`` (which is exactly what moveto is about)
- JS errors when calling ``WebDriver\Session::source()`` in some cases (meaning an issue in Selenium/GeckDriver probably)
- buggy handling of popup windows (the GeckoDriver status page documents the popup window support as being only partially implemented)
- issue in cookie resetting with path (not sure where the issue belongs)
- failures from #239
- failures about window name tests (see #256)

For the non-W3C APIs, we have to wait a bit for now, as replacement for the Selenium low-level (and badly documented) APIs we are using has been discussed only recently in the W3C WebDriver spec, and Firefox has not yet implemented this part.

aik099:
> failing tests are not related to the element being clicked out of view, but to other unimplemented APIs:

Yes, I know. But if somebody is to run these tests locally on a driver, that doesn't implement `moveto` for element outside of viewport, then we need to make sure, that just clicking would also pass the test.

stof:
@aik099 the W3C Webdriver spec says that ``Element::click`` must handle the scrolling part internally. So there is 3 solutions here:

- it is an implementation of the Selenium WebDriver protocol, and then moveto is part of it
- it is an implementation of the W3C protocol and then moving is useless
- it is an implementation of an unknown protocol or a buggy implementation of the above => we cannot know what it is the wau to support it.

The only thing I would do here is adding a test in the driver testsuite with a huge Lorem ipsum text and a link below it, so that the link is out of view for almost all screen sizes, to ensure that we can click on it through Mink.

> But if somebody is to run these tests locally on a driver, that doesn't implement moveto for element outside of viewport, then we need to make sure, that just clicking would also pass the test.

GeckoDriver is such a driver (as it implements the W3C protocol). And clicking works in it with this patch. This is why 35 tests changed status with this patch (33 from error to passing, and 2 form error to failing)

aik099:
Then why `moveto` was there in the first place if test should have worked just fine without `moveto` even on Selenium 2?

stof:
@aik099 Selenium2 is about the Selenium protocol, not the W3C protocol (Selenium 2 was written years before the standardization effort started at the W3C). and in the Selenium version of the protocol, moving into the view is not implied by ``Element::click`` (but moveto is available for that)

aik099:
I see now. I wasn't aware, that `moveto` became part of `click` only in Selenium 3.

stof:
@aik099 to be clear, Selenium 3 can run in both Selenium and W3C mode depending on the driver being used. The ChromeDriver is still in Selenium mode for instance

AmericanUmlaut:
@stof and @aik099  Hi. I'm blocked in a project that uses this library after an upgrade to Selenium 3 due to this issue with moveto in GeckoDriver. I would be very happy if this PR could be accepted - from reading the conversation it sounds like all that's still possibly required would be the test that stof suggests above?

Would it be possible to accept this PR without that test? If not, I'd be happy to develop the test myself if that would help get the wheels turning on this issue.

PR #256, opened 2017-01-03
Fix the window name tests for GeckoDriver

stof:
GeckoDriver does not put a UUID in its window names.

### DIFF
--- a/tests/Custom/WindowNameTest.php
+++ b/tests/Custom/WindowNameTest.php
@@ -6,7 +6,7 @@ use Behat\Mink\Tests\Driver\TestCase;
 
 class WindowNameTest extends TestCase
 {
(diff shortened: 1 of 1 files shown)


stof:
This removes 2 test failures on GeckoDriver.

However, I don't like my fix at all, as drivers are free to do what they want for their window name. AFAIK, the client code is meant to consider them as opaque strings. We should probably change these tests to remove the usage of this regex entirely. What do you think @aik099 ?

aik099:
When these tests were created I thought that window name should really obey that format. Later PhantomJS came in and decided to put it's own stuff in there.

Since window naming isn't a standard at all, then it would be best to just assert that different windows are given different non-empty window names, that we later can use to access them.

stof:
And btw, this should probably go in the driver-testsuite itself, as it was added as a Session-level API, not as a driver-specific API

aik099:
Probably.

PR #261, opened 2017-01-04
Fix the usage of the composer cache on Travis

stof:
Forcing a source install made the persistence of the composer download cache between builds useless, as we don't download packages.

No other (maintained) Mink repository is using ``--prefer-source`` on Travis. This one should not either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
 before_script:
   - sh bin/run-"$WEBDRIVER".sh
 
-  - composer install --prefer-source
+  - composer install
 
   # Start a webserver for web fixtures. Force using PHP 5.6 to be able to run it on PHP 5.3 and HHVM jobs too
   - MINK_PHP_BIN=~/.phpenv/versions/5.6/bin/php vendor/bin/mink-test-server > /dev/null 2>&1 &


aik099:
Shouldn't you use `--prefer-dist` option then? It even works if `dev-master` branch is used.

stof:
Well, only Mink and the driver-testsuite are installed using dev versions currently. I'm not sure it is worth keeping downloads for dev versions in the composer cache on Travis.
and this matches what all other repos are doing.

aik099:
Still it takes 17 seconds (I've restarted the build to see how cache is used) to install all dependencies. Probably the longest part isn't cloning of driver test suite, but composer checking latest version of each dependency.

This time can be dramatically shortened if `composer.lock` was committed.

stof:
> This time can be dramatically shortened if composer.lock was committed.

This is unmaintainable:

- we would not be testing the latest Mink and driver testsuite (as the lock file will not be updated automatically), meaning we would not test what users are really using
- it would not work when wanting to change deps in the matrix (testing against different versions of dependencies)

I'm strongly against committing the lock file for libraries (as the lock file would be meaningful only for the CI setup and not for real usage of the library).

> Still it takes 17 seconds (I've restarted the build to see how cache is used) to install all dependencies.

The longest time is probably the resolution time, and we cannot optimize it much here. The best optimization is to use PHP 7 (PHP 7.1 takes 12s with an empty cache)

aik099:
OK.

Issue #262, opened 2017-01-30
Relying on old package


uuf6429:
This package depends on `instaclick/php-webdriver`, whose last commit was 2 years ago.

Seems to me that `facebook/php-webdriver` is better supported... why not switch to this package? Are there any functional differences between the two?

aik099:
There was an attempt to do it in #79 PR, but nothing was merged due PR author seem to have abandoned it.

If you can switch to Facebook version of the Web Driver client without introducing a BC breaks, then why not.

uuf6429:
OK, cool. I'll give it a try.

smuggli:
Would love to see that! Worked with both drivers and facebook's is way better. If BCs come up, why not introduce a new mink driver.

uuf6429:
Guys, I rewrote it with my (minimal) understanding of this whole system:
https://github.com/minkphp/MinkSelenium2Driver/compare/master...uuf6429:Switch-to-FB-webdriver
However as you can see, something is broken:
https://travis-ci.org/uuf6429/MinkSelenium2Driver/builds/198412216
Unfortunately, I was not able to run the tests locally. I tried: running tests on my host (Windows) machine, running in [Ubuntu via WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux) and running via docker. All yielded the same failure, more or less (failed at setDesiredCapabilities test).

aik099:
The mentioned Travis CI build doesn't really help, because all tests end up with `The job exceeded the maximum time limit for jobs, and has been terminated.` message and don't show why they've failed.

aik099:
> I've removed testing on PHP 5.3/5.4 and bumped minimum PHP version to 5.5

Why?

> I've added PHPUnit to require-dev

Why?

aik099:
You'd better send a PR so that we can review it.

aik099:
And I already see many changes, that should be avoided, like:

* changing `array( ... )` into `[]` (the PHP 5.4 array syntax) - we must be PHP 5.3 compatible
* removing formatting from multi-line array declarations
* adding spaces around `.` (concatenation operator)
* whole file upload code was removed, but it was there to support remote file uploads; not sure if `sendKeys` in Facebook driver version is clever enough to do this
* code for setting desired capabilities was removed and probably that's why tests for it fail

uuf6429:
@aik099 
[...]

PR #263, opened 2017-02-05
Switch to fb webdriver

uuf6429:


.travis.yml
```diff
@@ -6,7 +6,7 @@ cache:
   directories:
     - $HOME/.composer/cache/files
 
-php: [5.3, 5.4, 5.5, 5.6, 7.0, 7.1, hhvm]
```
aik099: Revert. See https://github.com/minkphp/MinkSelenium2Driver/issues/262#issuecomment-277508479.

.travis.yml
```diff
@@ -35,7 +35,7 @@ before_script:
   # Start a webserver for web fixtures. Force using PHP 5.6 to be able to run it on PHP 5.3 and HHVM jobs too
   - MINK_PHP_BIN=~/.phpenv/versions/5.6/bin/php vendor/bin/mink-test-server > /dev/null 2>&1 &
 
-script: phpunit -v --coverage-clover=coverage.clover
```
aik099: Revert. See https://github.com/minkphp/MinkSelenium2Driver/issues/262#issuecomment-277508548.

composer.json
```diff
@@ -20,13 +20,14 @@
     ],
 
     "require": {
-        "php":                       ">=5.3.1",
```
aik099: Revert. See https://github.com/minkphp/MinkSelenium2Driver/issues/262#issuecomment-277508479.

composer.json
```diff
@@ -20,13 +20,14 @@
...
-        "behat/mink":                "~1.7@dev",
-        "instaclick/php-webdriver":  "~1.1"
+        "php":                   ">=5.5",
+        "behat/mink":            "~1.7@dev",
+        "facebook/webdriver":    "~1.3"
     },
 
     "require-dev": {
-        "mink/driver-testsuite": "dev-master"
```
aik099: Revert. See https://github.com/minkphp/MinkSelenium2Driver/issues/262#issuecomment-277508548.

phpunit.xml.dist
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<phpunit colors="true" bootstrap="vendor/autoload.php">
```
aik099: Revert. You can specify it for now in `.travis.yml` and command line option to PHPUnit. No need to alter default phpunit config file.

src/Selenium2Driver.php
```diff
@@ -68,25 +70,31 @@ class Selenium2Driver extends CoreDriver
...
+     * @param string $browserName Browser name
+     * @param array $desiredCapabilities The desired capabilities
+     * @param string $wdHost The WebDriver host
      */
     public function __construct($browserName = 'firefox', $desiredCapabilities = null, $wdHost = 'http://localhost:4444/wd/hub')
     {
         $this->setBrowserName($browserName);
         $this->setDesiredCapabilities($desiredCapabilities);
-        $this->setWebDriver(new WebDriver($wdHost));
+        $this->wdHost = $wdHost;
```
aik099: Any reason why we can't create WebDriver at this point as it was done before?

src/Selenium2Driver.php
```diff
@@ -68,25 +70,31 @@ class Selenium2Driver extends CoreDriver
...
     /**
      * Sets the browser name
      *
      * @param string $browserName the name of the browser to start, default is 'firefox'
+     *
+     * @throws DriverException
      */
     protected function setBrowserName($browserName = 'firefox')
     {
+        if ($this->started) {
```
aik099: How can somebody call protected method on a non-started session?

src/Selenium2Driver.php
```diff
@@ -107,45 +115,12 @@ public function setDesiredCapabilities($desiredCapabilities = null)
         }
 
         if (null === $desiredCapabilities) {
-            $desiredCapabilities = array();
```
aik099: Revert. Please don't use PHP syntax not compatible with PHP 5.3.

src/Selenium2Driver.php
```diff
@@ -107,45 +115,12 @@ public function setDesiredCapabilities($desiredCapabilities = null)
...
+            $desiredCapabilities = [];
         }
 
         // Join $desiredCapabilities with defaultCapabilities
         $desiredCapabilities = array_replace(self::getDefaultCapabilities(), $desiredCapabilities);
 
-        if (isset($desiredCapabilities['firefox'])) {
```
aik099: Any reasons for removing all that code? Maybe this is done differently now and if it is, then does related test pass?

src/Selenium2Driver.php
```diff
@@ -160,19 +135,9 @@ public function getDesiredCapabilities()
     }
 
     /**
-     * Sets the WebDriver instance
-     *
-     * @param WebDriver $webDriver An instance of the WebDriver class
-     */
-    public function setWebDriver(WebDriver $webDriver)
```
aik099: Maybe we can still allow that behavior somehow?

src/Selenium2Driver.php
```diff
@@ -186,17 +151,17 @@ public function getWebDriverSession()
      */
     public static function getDefaultCapabilities()
     {
-        return array(
-            'browserName'       => 'firefox',
```
aik099: No need to remove formatting of multi-line array calls.

src/Selenium2Driver.php
```diff
@@ -240,27 +205,27 @@ protected static function charToOptions($char, $modifier = null)
...
         if ($modifier) {
-            $options[$modifier.'Key'] = 1;
+            $options[$modifier . 'Key'] = 1;
         }
 
         return json_encode($options);
     }
 
     /**
-     * Executes JS on a given element - pass in a js script string and {{ELEMENT}} will
```
aik099: Why the `{{ELEMENT}}` was replaced with `__ELEMENT__`? If I recall correctly it's driver itself who parses `{{ELEMENT}}` and not WebDriver/Facebook.

src/Selenium2Driver.php
```diff
@@ -298,15 +263,37 @@ private function executeJsOnElement(Element $element, $script, $sync = true)
...
      * {@inheritdoc}
      */
     public function start()
     {
         try {
-            $this->wdSession = $this->webDriver->session($this->browserName, $this->desiredCapabilities);
+            $this->wdSession = new Selenium2DriverSession($this->buildWebDriver());
             $this->applyTimeouts();
         } catch (\Exception $e) {
-            throw new DriverException('Could not open connection: '.$e->getMessage(), 0, $e);
```
aik099: Revert. No need to change formatting.

src/Selenium2Driver.php
```diff
@@ -762,9 +752,13 @@ public function click($xpath)
         $this->clickOnElement($this->findElement($xpath));
     }
 
+    /**
+     * @param Element $element
+     */
     private function clickOnElement(Element $element)
     {
-        $this->wdSession->moveto(array('element' => $element->getID()));
+        //$element->click(); // TODO why was this needed?
```
aik099: What this comment is for? The `click` is called because method is called `clickOnElement` 😄

src/Selenium2Driver.php
```diff
@@ -793,36 +787,23 @@ public function attachFile($xpath, $path)
...
-        // as a remote instance.
-        try {
-          $remotePath = $this->uploadFile($path);
-        } catch (\Exception $e) {
-          // File could not be uploaded to remote instance. Use the local path.
-          $remotePath = $path;
-        }
-
-        $element->postValue(array('value' => array($remotePath)));
+        $element->sendKeys($path);
```
aik099: Are you sure, that `sendKeys` is clever enough to handle file upload code on local/remote servers?

phpunit.xml.dist
```diff
@@ -1,6 +1,6 @@
...
+<phpunit colors="true" bootstrap="vendor/autoload.php" stopOnFailure="true">
```
uuf6429: This is temporary until we get all tests working.

src/Selenium2Driver.php
uuf6429: Seems that the facebook webdriver already does this process... I have to be sure though.

src/Selenium2Driver.php
uuf6429: 1. It shows up as JS errors in my IDE
2. It seems to be replaced by webdriver: https://github.com/uuf6429/MinkSelenium2Driver/blob/05872053bd733959b25fa4df3f3aa41bef845d64/src/Selenium2Driver.php#L251-L251

src/Selenium2Driver.php
uuf6429: Is this behaviour really required? To me it seems like the perfect invitation for hackeries and seriously weird bugs.

src/Selenium2Driver.php
uuf6429: Ah, good point... didn't realise it's protected.

phpunit.xml.dist
aik099: Still please move it to `.travis.yml`. This would allow people checking out PR branch (for reviewing purposes) like me and using alternative PHPUnit runner not have this stopOnFailure behavior.

src/Selenium2Driver.php
uuf6429: If you look at the code of `buildWebdriver()`, you'll see that when the web driver is build, it requires the capabilities.
If I move that code up here, `setDesiredCapabilities()` will not work anymore.

src/Selenium2Driver.php
aik099: I'm not sure that we have tests to cover passing of browser-specific desired capabilities though. If we do, then they'll catch fact that something isn't passed.

src/Selenium2Driver.php
uuf6429: Yes: https://github.com/facebook/php-webdriver/blob/85aa68020ef2cf4fd261b1f1a690052afe67a9e0/lib/Remote/RemoteWebElement.php#L324

src/Selenium2Driver.php
aik099: Looks like that. I'll we'll address that later on.

src/Selenium2Driver.php
aik099: That will do for now then. Thanks for explaining.

src/Selenium2Driver.php
uuf6429: It's called `clickOnElement()`, not `doubleClickOnElement()`... `$element->click()` was called twice. What's the reason?

uuf6429: Ah, never mind... seems to be an artefact from my changes.

src/Selenium2Driver.php
uuf6429: Actually, I just realized... the point of a protected method is that it can be called in a child class, no?
So if someone calls it from child class, then the validation is still valid. If not, then we don't need this method in the first place and we can as well assign the property in the constructor.

src/Selenium2Driver.php
aik099: Of course it will show as JavaScript error in IDE because `{{ELEMENT}}` is invalid syntax. Since it works please replace `__ELEMENT__` back into `{{ELEMENT}}`.

src/Selenium2Driver.php
aik099: Fact, that no exception was thrown before proves that nobody really does that.

If you'd like, then you can move exception throwing into separate PR since it's not related to switching to Facebook's WebDriver in any way.

src/Selenium2Driver.php
aik099: Good. In any case we have functional tests to prove that it really works as expected.

src/Selenium2Driver.php
aik099: Then you can safely remove it.

src/Selenium2Driver.php
uuf6429: Why do you want to show errors when they're in fact just replaced tokens? O.o

aik099: Because changes in PR should be on topic. Making unrelated change doesn't help anybody.

aik099: You can send another PR with just that change and we can merge/not merge it separately regardless of faith of this PR.

stof: Plus, it is a BC break as it is a change affecting a protected method

stof:
Why closing the PR ?

aik099:
> Why closing the PR ?

Because nobody noticed upfront that Facebook's version of WebDriver requires PHP 5.5+ to run. Since Mink/Mink Driver require PHP 5.3+ to run we have no way to use it.

stof:
Well, as PHP 5.3 and 5.4 are EOLed, dropping support for them could be acceptable if we have some benefit when doing it (switching to a maintained library).
The previous PR dropping it for Mink was dropping support with no benefit for us. Here, it could have one.

Some drivers may drop support for older PHP versions (the third-party phantomJS driver does not support PHP 5.3 btw)

src/Selenium2DriverSession.php
```diff
@@ -0,0 +1,286 @@
...
+
+namespace Behat\Mink\Driver;
+
+use Behat\Mink\Exception\DriverException;
+use Facebook\WebDriver\Remote\DriverCommand;
+use Facebook\WebDriver\Remote\RemoteWebDriver;
+use Facebook\WebDriver\WebDriverBy;
+use Facebook\WebDriver\WebDriverElement;
+
+class Selenium2DriverSession
```
stof: Why creating this class ?

aik099:
> Here, it could have one.

Switching one dependency for another, while both work at the end is no benefit to me. Fact that one dependency is better maintained than another one doesn't really change the fact that both work.

src/Selenium2DriverSession.php
uuf6429: It's a proxy class to avoid changing the original driver too much. In the end, I could merge all changes into one class.

bwowk:
@aik099 actually there are many limitations on the instaclick version that are not present on the facebook one. Off the top of my head I can remember the iframe manipulation bit, for instance. On instaclick/php-webdriver you can only change iframes based on it's id, whereas in facebook/php-webdriver you can use the id, name or an WebDriverElement

aik099:
> Off the top of my head I can remember the iframe manipulation bit, for instance.

This particular change isn't a problem, because public API of Mink supporting locating frames by name only anyway.

Since Instaclick version covers 100% of functionality needed to implement the driver it doesn't really matter at this point if we switch to FB at all.

uuf6429:
Some other reasons:
- maintained library is possibly faster and in case of performance/bug/security issues, there's a higher chance of a patch
- the FB library takes care of things like handling uploads and other stuff that is unnecessarily replicated in MinkSelenium2Driver.

That said, I gave up arguing here. It's a huge shame this project will die like this "because it works".

bwowk:
@uuf6429 no reason for it to die at all, just put your fork up on Packagist and I'll help you maintain it. No trouble at all if the main one is stagnant and becomes obsolete, we'll just move forward using yours.

aik099:
I see you're eager to change stuff even though it works as expected for years now. I'm not here to judge how you should spend your time, but in your place I won't be spending time on doing anything that won't result in any kind of gain.

In this case I don't see any actual gains except `FB library is more maintained than Instaclick` one. There can't be any possible performance gain, because both FB and Instaclick are doing remote calls using WebDriver protocol, where speed is result of how fast we get a response and not how fast library can process it.

Einenlum:
Seems instaclick does not officially support Selenium 3.x, am I right?
Therefore, if a library is still active (facebook) and the other is not, plus does not supports officially Selenium 3.x, why shouldn't we change this dependency?

Moreover, if it's a question of BC and minimum PHP version, it's also our duty to make all this community evolve. Even 5.6 is not actively supported anymore. People with previous PHP versions will still have older versions of MinkSelenium2Driver.

PR #270, opened 2017-05-30
only sleep in wait() if the result is not true

individual-it:
there is no need to sleep if the result is already true, so save some time

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -952,7 +952,7 @@ JS;
         $start = microtime(true);
         $end = $start + $timeout / 1000.0;
 
-        while (microtime(true) < $end && !$this->wdSession->execute(array('script' => $script, 'args' => array())))
+        while (!($result=$this->wdSession->execute(array('script' => $script, 'args' => array()))) && (microtime(true) < $end))
         {
             usleep(100000);
         }
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -952,12 +952,10 @@ JS;
         $start = microtime(true);
         $end = $start + $timeout / 1000.0;
 
-        do {
-            $result = $this->wdSession->execute(array('script' => $script, 'args' => array()));
-            if (!$result) {
-                usleep(100000);
-            }
-        } while (microtime(true) < $end && !$result);
+        while (microtime(true) < $end && !$this->wdSession->execute(array('script' => $script, 'args' => array())))
+        {
+            usleep(100000);
+        }
 
         return (bool) $result;
     }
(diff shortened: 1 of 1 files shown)


aik099:
This might actually be anti-optimization, because if cycle iteration count while you wait for `$result` to become `true` is >1, then you actually loose time by checking `$result` again inside cycle.

stof:
@aik099 if the result is `true`, the loop will exit

aik099:
> @aik099 if the result is true, the loop will exit

That is how it works currently. But what I'm saying that extra code to check same exit condition only doubles code run time which is spending more time then is saved by calling `usleep` extra time after $result is true.

individual-it:
So you suggesting the if statement will take longer than 1/10s 

On May 30, 2017 2:02:48 PM GMT+05:45, Alexander Obuhovich <notifications@github.com> wrote:
>> @aik099 if the result is true, the loop will exit
>
>That is how it works currently. But what I'm saying that extra code to
>check same exit condition only doubles code run time which is spending
>more time then is saved by calling `usleep` extra time after $result is
>true.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/minkphp/MinkSelenium2Driver/pull/270#issuecomment-304807654

-

aik099:
It depends on how soon condition will become true. If element doesn't appear on the page for certain time, then total time spent on that extra IF execution might be close (I haven't measured really) to sleep time.

I think you'd better attach some benchmarks for cases with extra IF and without it with different waiting times (e.g. element appeared after X milliseconds, sleep time, extra if time, total time).

individual-it:
What about changing it around and having a while and not a do...while loop? Then there are the same amount of checks, but the loop only waits if the condition is not true. If the condition returns true strait away we save 1/10s

src/Selenium2Driver.php
```diff
@@ -952,10 +952,10 @@ public function wait($timeout, $condition)
         $start = microtime(true);
         $end = $start + $timeout / 1000.0;
 
-        do {
-            $result = $this->wdSession->execute(array('script' => $script, 'args' => array()));
+        while (microtime(true) < $end && !$this->wdSession->execute(array('script' => $script, 'args' => array())))
```
stof: this does not ensure that the execution runs at least once

individual-it: the only case I can imagine in which it would not run at all is when `microtime(true) < $end` in the very first iteration of the loop. That would happen if the user specify a timeout that is shorter than the time needed for setting up the while loop and test `microtime(true) < $end` but that is fine because in that case the timeout is actually run up.

But I also could reverse the loop
`while (!$this->wdSession->execute(array('script' => $script, 'args' => array())) && microtime(true) < $end)`
Then the JS will run always at least once, no matter of the timeout

individual-it:
sorry $result need to be set of course

PR #273, opened 2017-10-06
See what happens in Travis

phil-davis:
then see how it could be fixed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,9 @@ before_script:
   - composer install
 
   # Start a webserver for web fixtures. Force using PHP 5.6 to be able to run it on PHP 5.3 and HHVM jobs too
-  - MINK_PHP_BIN=~/.phpenv/versions/5.6/bin/php vendor/bin/mink-test-server > /dev/null 2>&1 &
+  - ls -l ~/.phpenv/versions/5.6/bin/php
+  - if [[ ${TRAVIS_PHP_VERSION:0:1} == "5" ]]; then MINK_PHP_BIN=~/.phpenv/versions/5.6/bin/php vendor/bin/mink-test-server > /dev/null 2>&1 & ; fi
+  - if [[ ${TRAVIS_PHP_VERSION:0:1} == "7" ]]; then vendor/bin/mink-test-server > /dev/null 2>&1 & ; fi
 
 script: phpunit -v --coverage-clover=coverage.clover
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,8 @@ before_script:
 
   # Start a webserver for web fixtures. Force using PHP 5.6 to be able to run it on PHP 5.3 and HHVM jobs too
   - ls -l ~/.phpenv/versions/5.6/bin/php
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+1.3.2 / 2017-10-09
+==================
+
+Bug fixes:
+
+* See what goes wrong in Travis then fix it
```
stof: WTF ? Why adding such thing in the changelog (and as a bugfix)

.travis.yml
```diff
@@ -33,7 +35,13 @@ before_script:
   - composer install
 
   # Start a webserver for web fixtures. Force using PHP 5.6 to be able to run it on PHP 5.3 and HHVM jobs too
-  - MINK_PHP_BIN=~/.phpenv/versions/5.6/bin/php vendor/bin/mink-test-server > /dev/null 2>&1 &
+  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.3" ]]; then MINK_PHP_BIN=~/.phpenv/versions/5.3/bin/php vendor/bin/mink-test-server > /dev/null 2>&1 & fi
```
stof: the mink test server cannot run on 5.3 (see the comment above)

phil-davis: As you can see, I now have no idea what to do next, but am trying various different stuff to see what might be going on. The release of any new version of MinkSelenium2Driver seems to be stalled because for some reason the testing CI fails. If I could work out what the problem is and fix it, then happy to contribute.

phil-davis:
I have no idea of the history of this. For somebody who knows what bits go with what in which versions of PHP, phpunit, webdrivers... is getting the tests to pass likely to be a "short" exercise?

Or is there some architectural issue between all the components that is somehow extremely difficult to fix?

The next official release has been waiting for a long time. Other than me hacking around and not knowing what to do next, is there something that can be done to help get the tests to pass and a new release out? Is it a human resource issue, or some deep technical (read almost unfixable) issue?

stof:
Well, the main reason is human resource.

I actually have a branch locally with some changes to help regarding tests (I started doing this work a few days ago, but without having the time to finish it)

phil-davis:
I am happy to contribute time, but the problem is that my time is useless because I am starting from scratch and having to dig around by myself in every message I see in the Travis log. 
e.g. in PHP 5.* jobs:
```
PHP Fatal error:  Call to a member function window_handles() on a non-object in /home/travis/build/minkphp/MinkSelenium2Driver/src/Selenium2Driver.php on line 492
```
It seems that ``wdSession`` never gets set, and I guess there is some problem in getting the webdriver started, connected or...

In PHP 7.* jobs:
```
Fatal error: Uncaught Error: Class 'PHPUnit_Runner_Version' not found in /home/travis/build/minkphp/MinkSelenium2Driver/vendor/mink/driver-testsuite/tests/SkippingUnsupportedTestCase.php on line 7
```
I guess I need to set ``MINK_PHP_BIN`` or something in the environment?

So it is not practical to bring in someone new to do this sort of job.

PR #277, opened 2018-01-07
Clean default capabilities

stof:
Setting capabilities specifying a firefox version or a selenium version by default breaks when running against a server being strict about the requested capabilities (as it is unlikely to run Firefox 9 these days).
Note that I faced this issue when trying to run tests against geckodriver (as it is strict about respecting the Firefox version, unlike the old FirefoxDriver of Selenium 2).

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -189,14 +189,7 @@ class Selenium2Driver extends CoreDriver
     {
         return array(
             'browserName'       => 'firefox',
-            'version'           => '9',
-            'platform'          => 'ANY',
-            'browserVersion'    => '9',
-            'browser'           => 'firefox',
             'name'              => 'Behat Test',
-            'deviceOrientation' => 'portrait',
-            'deviceType'        => 'tablet',
-            'selenium-version'  => '2.31.0'
         );
     }
 


usmonster:
...I guess I can close #274, then?

Also, I believe the default `version` might need to be the empty string, which should represent the "latest" version of the browser. I think it's a better default than forcing users to specify version explicitly all the time (and it's backwards compatible). What do you think?

src/Selenium2Driver.php
```diff
@@ -189,14 +189,7 @@ public static function getDefaultCapabilities()
     {
         return array(
             'browserName'       => 'firefox',
-            'version'           => '9',
```
usmonster: Can we make the default `version` the empty string, which signifies the "latest" available version of the browser. I think it's a better default than forcing users to specify the browser version explicitly (and it would make the change backwards compatible). What do you think?

src/Selenium2Driver.php
```diff
@@ -189,14 +189,7 @@ public static function getDefaultCapabilities()
...
-            'platform'          => 'ANY',
```
usmonster: Following similar reasoning as my previous comment, can this capability also be left in? This seems a reasonable default, and it would be convenient and intuitive to allow users to leave it out of their own capabilities when they don't care. Thoughts?

src/Selenium2Driver.php
usmonster: Can we make the default `version` the empty string, which signifies the "latest" available version of the browser (or, technically, [an "unknown" version](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities#used-by-the-selenium-server-for-browser-selection)). I think it's a better default than forcing users to specify the browser version explicitly, and it would make the change backwards compatible.

What do you think?

stof: is it actually mandatory to provide a version in the capabilities ? In my testing, not passing the capability does the same than allowing any value in it.

usmonster: In some versions of Selenium, at least, this is a required parameter. It might be safer to pass the empty string, as documented in the link I mentioned above.

src/Selenium2Driver.php
usmonster: @stof Any response to this? `'ANY'` is likely needed if we don't want to force the user to explicitly specify the platform when it doesn't matter.

Issue #287, opened 2018-06-25
Fixing the Build


aik099:
All of recent builds are failing due different reasons.

- The PHP 5.3 build fails due incorrect Ubuntu version usage (no PHP 5.3 available on it)
- The PHP 5.x builds fail due Fatal Error (because Selenium 2 doesn't work on latest Firefox provided by Travis CI)
- The PHP 7.x builds fail (almost all tests failed) due timeout issues connecting to Selenium Server (have no idea why)
- The PHP 7.x build using remote virtual machine for running selenium works, but have 2 tests failed

aik099:
Currently I'm stuck on https://github.com/minkphp/driver-testsuite/issues/21, when attempting to run tests locally.

jibran:
> Selenium 2 doesn't work on latest Firefox provided by Travis CI

I think this can be fixed by https://github.com/instaclick/php-webdriver/commit/be87042314cc654f7ff3e26c20a998cadb23daac. PR for updating the library is https://github.com/minkphp/MinkSelenium2Driver/pull/289 but the larger discussion is https://github.com/minkphp/MinkSelenium2Driver/issues/254 which is of course block on this. :slightly_smiling_face:

aik099:
Maybe. I was able to get test suite working (at least not a fatal error) on PHP 5.5 locally using Chrome. I've attempted to get same effect on Travis CI using #290, but it's not working. Attempt to debug the build also failed with `access denied` error.

greg0ire:
How about just dropping php 5 support, if fixing php 5 builds is all that is left to do?

aik099:
@greg0ire , there are multiple similar issues. There are also PRs created, that use Facebook Driver instead of InstaClick, that also require PHP 7 at least.

greg0ire:
Similar issues about dropping php 5? Or something else? I can create a PR to drop php 5 if you want me to, I think doing so in a minor release would not be an issue since php 5 is EOL'ed: https://www.php.net/supported-versions.php

aik099:
Issues/PRs are about switching to different WebDriver implementation. It just happens, that different WebDriver implementation require PHP 7 at minimum and that would in turn force to increase minimal supported version for this repository as well.

greg0ire:
[...]

PR #289, opened 2018-07-01
Support PHP WebDriver 2.0.

### LINKED ISSUE #288: Support PHP WebDriver 2.0
See https://github.com/instaclick/php-webdriver/releases/tag/2.0.0-BETA

jibran:
Fixes #288

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "require": {
         "php":                       ">=5.3.1",
         "behat/mink":                "~1.7@dev",
-        "instaclick/php-webdriver":  "~1.1"
+        "instaclick/php-webdriver":  "~1.1|~2.0@RC"
     },
 
     "require-dev": {


aik099:
Without fixed builds I can't really see any difference that new WebDriver library version does. 

We should really focus on #287 first and then try adding more features.

jibran:
> I can't really see any difference that new WebDriver library version does.

I think it will fix #254 but yeah, it is not going to be straightforward. Maybe we have to bump major version or add BC layer.

Techwolf12:
For #254, it is not fuly compatible by just including this:
    And I set browser window size to 1440 x 1024                               # FeatureMinkContext::iSetBrowserWindowSizeToX()
      Notice: Undefined index: size in vendor/instaclick/php-webdriver/lib/WebDriver/AbstractWebDriver.php line 204

I'll check it out

PR #294, opened 2018-10-03
Rewrite the window name tests to avoid forcing a pattern

stof:
Selenium defines the window name as a opaque string, so each browser is free to use anything they want. And this changed over time, causing the UUID regex to break even for browsers using a UUID in the past (Chrome adds more things around the UUID for instance).


aik099:
OK.

stof:
this makes these tests pass when running them locally against chrome and firefox

PR #295, opened 2018-10-03
Fix some Travis issues

stof:
- remove testing on HHVM, as they stopped caring about PHP compat
- use the precise infra to run tests on PHP 5.3 (until we drop support, which is not the goal here)
- add testing on PHP 7.2
- use the tested PHP version to run the builtin webserver, as the Trusty infra may not pre-install PHP 5.6 (except for the 5.3 job of course, but that one runs on Precise)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ cache:
   directories:
     - $HOME/.composer/cache/files
 
-php: [5.3, 5.4, 5.5, 5.6, 7.0, 7.1, hhvm]
+php: [5.4, 5.5, 5.6, 7.0, 7.1, 7.2]
 
 env:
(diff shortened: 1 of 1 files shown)


aik099:
What build errors exactly are fixed by this PR? Can you please quote the covered Travis CI error messages? I'm asking, because you might be fixing errors reported in other issues as well.

stof:
The second point in my list fixes the fact that the PHP 5.3 job is crashing saying it cannot have PHP 5.3 (https://travis-ci.org/minkphp/MinkSelenium2Driver/jobs/436585583)

The other points are cleanups of the test infra (and adding PHP 7.2).

aik099:
OK.

PR #296, opened 2018-10-03
Fix the test about setting invalid timeout

stof:
If the session is not started, the timeout is not applied on the webdriver session (as it does not exist at that point). Failures happen only when applying the timeout on the session.

That's a test I missed in #278 when fixing other tests which were dependant on the implicit start during `getSession()`


stof:
Note that this test was failing only when being run as the first test of the testsuite, as our testsuite is resetting the session between tests, not stopping it.

aik099:
OK.

PR #297, opened 2018-10-03
Improve the timeout handling

stof:
The first commit improves the isolation for timeout tests, to avoid impacting other tests (this is especially noticeable when they break, as they were breaking next tests as well by keeping the timeouts)

The second commit improves the way we handle the started state vs applying timeouts, to avoid leaking webdriver sessions (which would not close the browser).

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -298,15 +298,13 @@ class Selenium2Driver extends CoreDriver
     {
         try {
             $this->wdSession = $this->webDriver->session($this->browserName, $this->desiredCapabilities);
-            $this->applyTimeouts();
         } catch (\Exception $e) {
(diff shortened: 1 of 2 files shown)


aik099:
I wonder how you've noticed, that something timeout-related is even happening.

stof:
Trying to run such these tests in Firefox, which fails in weird ways due to using the W3C protocol which is different for timeouts, and then seeing the same failures on next tests *not* using timeouts :smile:

Btw, this allowed me to report https://github.com/instaclick/php-webdriver/issues/88 and https://github.com/instaclick/php-webdriver/issues/87 by finding 2 issues in the handling of the W3C protocol, and seeing that the 2.0-RC version of the webdriver library was still having the same code.

PR #299, opened 2018-10-10
Fix the configuration of the PHP 5.3 job on precise

stof:
Env variables defined as matrix are not merged with the job env variables of the matrix. We need to use global env variables (which we can still override in a specific job if needed).

This fixes this error: https://travis-ci.org/minkphp/MinkSelenium2Driver/jobs/436675862#L167

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,8 @@ cache:
 php: [5.4, 5.5, 5.6, 7.0, 7.1, 7.2]
 
 env:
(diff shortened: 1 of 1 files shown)


stof:
To win time on CI, I cancelled all the other jobs failing due to trying to use Firefox through the JsonWire protocol. they would fail anyway.

stof:
Looks like the precise job has an issue running the test server, but this would have to be debugged separately.

PR #290, opened 2018-07-01
Attempts to fix currently failing test suite

aik099:
Closes #287

PR #300, opened 2018-10-10
Use chrome on Travis rather than Firefox for most jobs

stof:
On Trusty, the preinstalled Firefox version is newer than the latest version supported by the FirefoxDriver of Selenium 2, and requires using the W3C protocol. But we don't support this protocol yet.

Replaces #290

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,22 +11,24 @@ php: [5.4, 5.5, 5.6, 7.0, 7.1, 7.2]
 env:
   global:
     - WEBDRIVER=selenium
+    - WEB_FIXTURES_BROWSER=chrome
 
 matrix:
   fast_finish: true
   include:
     - php: 7.0
-      env: WEBDRIVER=selenium-remote
+      env: WEBDRIVER=selenium-remote WEB_FIXTURES_BROWSER=firefox
       sudo: required
       services:
         - docker
     - php: 5.3
       dist: precise
       # Force using PHP 5.6 for the test server as PHP 5.3 does not have the builtin webserver
-      env: MINK_PHP_BIN=~/.phpenv/versions/5.6/bin/php
+      # Use Selenium 2 and Firefox when running on the precise infra, as it has a version of Firefox old enough.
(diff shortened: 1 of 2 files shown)


stof:
This also uses Selenium 3 rather than Selenium 2 for most jobs, as Selenium 2 is not maintained anymore.

aik099:
OK.

stof:
OK, chrome is crashing, but I have an idea why.

aik099:
Try using that scenario locally. Maybe you'll see what's actually happening, e.g. Chrome window opens and then closes after a specific test. Maybe we're attempting to connect before Chrome was able to start.

stof:
@aik099 I know what happens. Chrome's sandbox is enabled by default, but it is not compatible with Travis's container setup (containers are not priviledged enough to enable the sandboxing). And currently, Selenium2Driver does not auto-detect Travis to disable the chrome sandbox. It requires users to manually specify it.

We have 2 choices:

- adding the feature of automatically detecting Travis to disable the sandbox (if not done explicitly already)
- disabling it only in our testsuite.

what do you prefer ?

stof:
I will need to debug the build, as it fails many things and takes a very long time

aik099:
According to https://stackoverflow.com/questions/33013378/how-can-we-disable-web-security-of-chrome-browser-using-selenium-testng disabling sandboxing is done via capabilities during driver creation. If that is so, then we can:

1. create an option in `phpunit.xml.dist` for chrome sandboxing (not set by default = use whatever happens currently)
2. when creating a driver we pick that option and set correct capabilities
3. in the `travis.yml` we always disable sandboxing by packing `phpunit.xml.dist` file

stof:
Well, for now, I made a solution only for our own testsuite (so no impact on our users yet, which would be discussed separately once our CI works again), detecting Travis.

mvorisek:
This is 2 years old, I belive it can be closed now.

oleg-andreyev:
@mvorisek maybe it worth a look at https://github.com/oleg-andreyev/MinkPhpWebDriver

Issue #301, opened 2018-10-17
Method setValue in google places problem.


JordiGiros:
Some pages that use google places in their inputs delete their values when losing focus from the element. When the setValue method is called, his behaviour makes the input lose the focus and the place can't be selected because the value typed is lost. That makes some test to crash. 

The problem seems to be related to the "Prevent firing the "change" event twice" problem. 
Some time ago, the method used a TAB added after the value that was making the problem happen too.
It was solved [https://github.com/minkphp/MinkSelenium2Driver/pull/244](url) deleting the TAB but then the problem returned with [https://github.com/minkphp/MinkSelenium2Driver/pull/286](url).

After reading lots of problems related with this two behaviours of setValue method (causing problems in each type of solutions), why can a parameter be passed to the method in order to use is with or without losing focus?

Is there any alternative to set the value without losing the focus from input?
Thank you.

stof:
If we only type in the field without moving the focus away, there is *no* change event in the browser. That's why `setValue` moves the focus away.

JordiGiros:
https://www.hoppa.com/ie -> Drop off location has a clear example of what I'm telling. I can find these problem in some other Ryanair.com or Suntransfers booking process inputs.

JordiGiros:
www.suntransfers.com -> Booking process step 3 Drop-off point have the same problem.

JordiGiros:
https://www.airbnb.com/ -> Search "To" also closes the dropdown from google places but does not delete the value from the input. You have to "refocus" the search input field.

JordiGiros:
This issue seems to be duplicated with [https://github.com/minkphp/MinkSelenium2Driver/issues/292](url)

JordiGiros:
As I understund, it would be as easy as adding into Mink's CoreDriver.php
´´´
/**
     * {@inheritdoc}
     */
    public function setAutocompleteValue($xpath, $value)
    {
        throw new UnsupportedDriverActionException('Setting the field value is not supported by %s', $this);
    }

´´´

And DriverInterface.php
´´´
    /**
     * Sets autocomplete element's value by it's XPath query.
     *
     * @param string            $xpath
     * @param string|bool|array $value
     *
     * @throws UnsupportedDriverActionException When operation not supported by the driver
     * @throws DriverException                  When the operation cannot be done
     *
     * @see \Behat\Mink\Element\NodeElement::setAutocompleteValue
     */
    public function setAutocompleteValue($xpath, $value);
[...]

Issue #292, opened 2018-07-24
Blur at end of setValue causes StaleElementReference


phil-davis:
If ``fillField`` - ``setValue`` is called with an input that makes the field go away, then the recent code that tries to do a ``blur`` out of the field fails with ``StaleElementReference``

e.g. put an ENTER at the end of the field. In ownCloud, when creating a new folder, a dialog pops up and you type the new folder name and press enter. The dialog disappears and the new folder is created. See:
https://github.com/owncloud/core/issues/32143

After https://github.com/minkphp/MinkSelenium2Driver/pull/286 the above sequence fails with ``StaleElementReference`` because the new code in ``setValue`` tries to do a blur on the element after the data has been typed in the field, and the field has gone away.

It is possible to workaround by catching and ignoring the ``StaleElementReference`` exception. But it would be nice if the exception was not thrown in the first place.

phil-davis:
It is worse. We have text fields where you type in the field and do not press enter. The UI underneath fires off code to find matches of what you have type (matching user names, group names, file names, whatever). But ```setValue`` is removing focus from the field, so the underlying code to display the matches never triggers, and in any case we are unexpectedly taken away from the field.

Doing ``$shareWithField->focus();`` again does not help - it puts us back in the field at the end of the typed text, but then there needs to be another change to the value in the field to trigger the match code. If we type again in the field with ``setValue`` then it takes away the focus!!!

I can't think how to work-around that.

phil-davis:
Issue for workarounds in ownCloud: https://github.com/owncloud/core/issues/32143
PR for workarounds in ownCloud: https://github.com/owncloud/core/pull/32144
It might help someone else work around this.

aik099:
@phil-davis , that shouldn't be a problem if, in #286 discussion, we decide to return to original code prior to #244.

phil-davis:
If something can be done so that fillField/setValue ends up leaving focus on the field, then auto-complete has a chance...

aik099:
No `change` event is fired in real browser until focus leaves the field. This way no matter how we try we can't both support auto-complete controls and fire `change` event. Another point made in #286 is that no auto-complete control relies on `change` event. I guess it's likely that `blur` event happens on focus loss and that causes auto-complete to close.

JordiGiros:
@phil-davis Another option I've found on this is using javascript injection. 
[...]

PR #302, opened 2018-10-18
Add setAutocompleteValue implementing Mink DriverInterface

JordiGiros:
Implemented method "setAutocompleteValue" in order to solve setting a value on autocomplete input elements as commented in [286](https://github.com/minkphp/MinkSelenium2Driver/pull/286), [292](https://github.com/minkphp/MinkSelenium2Driver/issues/292) and  [301](https://github.com/minkphp/MinkSelenium2Driver/issues/301)

Related to [767](https://github.com/minkphp/Mink/pull/767) PR open in Mink project.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -670,8 +670,6 @@ JS;
 
         if (in_array($elementName, array('input', 'textarea'))) {
             $existingValueLength = strlen($element->attribute('value'));
-            // Add the TAB key to ensure we unfocus the field as browsers are triggering the change event only
-            // after leaving the field.
             $value = str_repeat(Key::BACKSPACE . Key::DELETE, $existingValueLength) . $value;
         }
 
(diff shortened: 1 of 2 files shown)


src/Selenium2Driver.php
```diff
@@ -692,6 +690,37 @@ public function setValue($xpath, $value)
         $this->executeJsOnElement($element, $script);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function sendKeys($xpath, $value)
```
aik099: Instead of duplication of code from “setValue” method plus few changes I’m suggesting to move shared code into new private method (or maybe protected one).

src/Selenium2Driver.php
```diff
@@ -692,6 +690,37 @@ public function setValue($xpath, $value)
...
+    {
+        $element = $this->findElement($xpath);
+        $elementName = strtolower($element->name());
+
+        if ('select' === $elementName) {
```
aik099: Why this IF is needed. Instead I'm proposing to allow only these elements in method:

* `input` HTML elements with specific `type` attribute values (already done)
* maybe `textarea` elements

aik099: Why this IF is needed. Instead I'm proposing to allow only these elements in method:

* `input` HTML elements with specific `type` attribute values (already done)
* maybe `textarea` elements

src/Selenium2Driver.php
```diff
@@ -1211,4 +1221,21 @@ private function uploadFile($path)
...
+     * @param $element
+     */
+    private function postElementValue($value, $elementName, $element)
+    {
+        $value = strval($value);
+
+        if (in_array($elementName, array('input', 'textarea'))) {
+            $existingValueLength = strlen($element->attribute('value'));
+            $value = str_repeat(Key::BACKSPACE . Key::DELETE, $existingValueLength) . $value;
+        }
```
aik099, line 1236: If I understood @stof comment (see https://github.com/minkphp/Mink/pull/767#issuecomment-431019669) correctly, then this should be moved back to `setValue` method to allow `sendKeys` to send keys (as the name suggests) instead of replacing current input value with provided one (as the `setValue` method does).

aik099 (changes requested):

PR #303, opened 2019-01-04
Fixing build

oleg-andreyev:
Closes #290 
Closes #287

Issue #298, opened 2018-10-03
[RFC] Name of the driver


stof:
Currently, the driver is named Selenium2Driver. But this driver is not really about Selenium 2 only. Currently, it is also about Selenium 3. And with #293, it would also be about the W3C Webdriver protocol, which is not related to Selenium itself (except for the inspiration).

In the new world, Selenium might not be involved at all in the stack now that Webdriver is a standard implemented by browsers directly, and Selenium is only the intermediate node for grid support (if we except Internet Explorer support, which is still provided by Selenium and might speak only the legacy protocol).

Should we have it as WebdriverDriver instead ?

What do you think @aik099 ?

stof:
Note that starting a new repo might be the easiest way to change the name in case we decide that the right path for #293 is to change our dependency.

aik099:
Shouldn't be `WebdriverDriver` be `WebDriverDriver` instead?

Otherwise I'm ok with a renaming. I see 3 ways:

1. rename repo on GitHub, but keep Packagist package name (keeping same name on Packagist is preserving BC, but defeating purpose of whole rename)
2. rename repo on GitHub and abandon current package in favor of new Packagist package (does that allow us to rename classes inside repo itself?)
3. having another repo + another package as if it was a new driver and advertising to use new package instead (would be a mess with currently opened issues/PRs, because whoever submits the issue/PR needs to figure out correct repo)

stof:
@aik099 if we rename it, the package and the code should be renamed too.

The strategy (2 or 3) will depend how different the code is with the facebook webdriver library, to see whether it is worth keeping the history, or whether starting from scratch is cleaner.

Strategy 2 allows to rename classes if we do a major version bump

aik099:
Then we're good to go. Not sure about currently opened issues/PRs. If we'll create another repo, then what will become of them?

stof:
>     3\. (would be a mess with currently opened issues/PRs, because whoever submits the issue/PR needs to figure out correct repo

I think having another repo for the enw package would be *less* of a mess. The new driver implementation would be a full rewrite, so all currently opened issues would be irrelevant for it.

stof:
Regarding the name, the W3C spec is indeed `WebDriver` and not `Webdriver`.

The only issue with that name is that `Mink\WebDriverDriver` and `mink/web-driver-driver` look a bit weird.

For the package name, maybe we can use `mink/webdriver-driver`.

stof:
[...]

Issue #293, opened 2018-09-24
Support the W3C webdriver spec


stof:
Support for the JsonWire selenium protocol is fading out in favor of the W3C spec:

- Firefox's geckodriver implements only the W3C protocol
- Safari's driver uses the W3C driver by default (see https://github.com/minkphp/Mink/issues/764)
- Chrome's driver is working on supporting the W3C protocol (they still implement JsonWire for now)
- Edge's driver uses the W3C driver by default in its next release (so same situation than Safari will happen soon: https://docs.microsoft.com/en-us/microsoft-edge/webdriver#webdriver-server-command-line-flags)

We will have some work to do to make this happen.

The instaclick/php-webdriver library that we are using has a beta version of 2.0 which adds support for interacting with a W3C server. But they won't provide an abstraction layer to make the same code work for both variants (see https://github.com/instaclick/php-webdriver/issues/84). It will be up to our code to choose the right way (and we will need https://github.com/instaclick/php-webdriver/issues/86 to be able to know which one we need).
Due to the fact that we expose the webdriver session to userland, we would have to consider this upgrade as a BC break.

Another option could be to migrate to facebook/webdriver, which seems a bit more popular this days (https://packagist.org/?query=webdriver). It is currently adding support for the W3C spec (WIP is at https://github.com/facebook/php-webdriver/pull/560), and does so by abstracting the differences internally).
Doing such migration would also require making it a major release, as we would change the API we expose when exposing the session.

stof:
https://github.com/minkphp/MinkSelenium2Driver/issues/288 was created to refer to the first solution for this topic (but does not describe the implication, and does not seem to care about the W3C support in the PR implementing it)

stof:
#287 should be done first before this work (we should switch to Chrome on Travis, as there is no way to run recent Firefox versions with the JsonWire protocol).

stof:
The more I look at instaclick/php-webdriver 2 "supporting" the  WEC protocol (currently in RC), the more I think we should rather migrate to the facebook package (for which the PR supporting the W3C protocol inside their abstraction is almost ready AFAICT).

stof:
[...]

PR #304, opened 2019-01-04
[WIP] Facebook driver

### LINKED ISSUE #97: Use native WebDriver event triggering, rather than Syn ones
We have several problems due usage of Syn library for JavaScript event triggering. I propose to switch to WebDriver native event triggering capabilities instead of using Syn library.

During the change we need to carefully check, that when using native WebDriver calls all JavaScript events, that were triggered by Syn library are still triggered by WebDriver itself.

Closes:
- #78 
- #75 
- #51 
- #20 
- #139

### LINKED ISSUE #175: Set desired capabilities for firefox
I couldn't add capability to firefox driver. I added proxy option for connection but I couldn't set 'permissions.default.image' as '2'. Briefly, I want to disable image loading while test continue. What am I doing wrong?

``` php
$driver = new \Behat\Mink\Driver\Selenium2Driver(
    'firefox',
    array(
        "proxy" => array(
            "proxyType" => "manual",
            "httpProxy" => Core::selectProxy(),
        ),
        "permissions.default.image" => 2

    )
);
```

(also I'm tried to send it as array)

### LINKED ISSUE #257: CI setup should be improved to test on Selenium 3
I plan to add 2 more jobs in the matrix:

- Selenium 3 using ChromeDriver (chromium is now available on Travis, so ChromeDriver is usable if we download it)
- Selenium 3 using GeckoDriver and a recent firefox version, as allowed failure (to track progress on it)

I'm wondering whether we should also add Selenium 2 + ChromeDriver in the matrix. not sure it is worth it.

oleg-andreyev:
**This is work in progress**

- Moving away from instaclick/php-webdriver to facebook/php-webdriver as first one is not supported (https://github.com/minkphp/MinkSelenium2Driver/issues/293#issuecomment-426680197)
- Removing syn.js
- Preparing to be more W3C compliant
- We should release a new major version of Mink, because facebook/php-webdriver needs PHP >= 5.6

Depending on #303, https://github.com/facebook/php-webdriver/pull/621, https://github.com/minkphp/driver-testsuite/pull/26, https://github.com/minkphp/driver-testsuite/pull/25
Unblocks #298, #293 
Closes #300 
Closes #289
Closes #257 
Closes #175
Closes #97


aik099:
There are other closed PRs doing the same, that might help you:

* #263
* #79

composer.json
```diff
@@ -1,46 +1,62 @@
...
-    "name":         "behat/mink-selenium2-driver",
-    "description":  "Selenium2 (WebDriver) driver for Mink framework",
-    "keywords":     ["selenium", "webdriver", "javascript", "ajax", "testing", "browser"],
-    "homepage":     "http://mink.behat.org/",
-    "type":         "mink-driver",
-    "license":      "MIT",
-
+    "name": "behat/mink-selenium2-driver",
+    "description": "Selenium2 (WebDriver) driver for Mink framework",
+    "keywords": [
```
aik099: If you want to do the reformatting of anything, then please create separate PR for that.

composer.json
```diff
@@ -1,46 +1,62 @@
...
+            "email": "pete@otaqui.com",
+            "homepage": "https://github.com/pete-otaqui"
+        },
+        {
+            "name": "Konstantin Kudryashov",
+            "email": "ever.zet@gmail.com",
+            "homepage": "http://everzet.com"
+        }
+    ],
+    "repositories": [
```
aik099: Surely this change isn't meant to be merged.

composer.json
```diff
@@ -1,46 +1,62 @@
...
+            "url": "git@github.com:oleg-andreyev/php-webdriver.git"
         }
     ],
-
     "require": {
-        "php":                       ">=5.3.1",
-        "behat/mink":                "~1.7@dev",
-        "instaclick/php-webdriver":  "~1.1"
+        "php": ">=5.3.1",
+        "ext-zip": "*",
```
aik099: All added `ext-` stuff should be a separate PR where you'll explain why it's needed.

composer.json
```diff
@@ -1,46 +1,62 @@
...
+        "ext-curl": "*",
+        "ext-json": "*",
+        "behat/mink": "~1.7@dev",
+        "facebook/webdriver": "dev-adjust-getRelatedElements"
```
aik099: You're not using a stable release. Why?

Building off an integration based on locally modified fork seems to be a bad idea.

composer.json
```diff
@@ -1,46 +1,62 @@
...
     },
-
     "require-dev": {
-        "mink/driver-testsuite": "dev-master"
+        "roave/security-advisories": "dev-master",
+        "mink/driver-testsuite": "dev-w3c-testKeyboardEvents",
```
aik099: Why? Isn't current test suite working fine?

composer.json
```diff
@@ -1,46 +1,62 @@
...
+        "phpunit/phpunit": "^6.5"
```
aik099: Why? Everything works pretty fine with globally installed PHPUnit.

composer.json
oleg-andreyev: right, this PR is currently WIP.

Why it's needed:
- facebook/php-webdriver has an issue with detecting related <form/> element (already submitted https://github.com/facebook/php-webdriver/pull/621)
- mink/driver-testsuite is required because of https://github.com/minkphp/driver-testsuite/pull/25 (facebook/php-webdriver is a bit more w3c complaint and w3c specifications does not have keyup/keydown specifications)

composer.json
oleg-andreyev: afaik it's reported by https://github.com/kalessil/phpinspectionsea plugin, 
I've removed it, as curl/zip/json aren't required for this lib

oleg-andreyev:
Green build 🥇

stof:
>     * We should release a new major version of Mink

not necessary. That's fine to have a different requirement in the new version of the driver.

We need a new major version of the driver though, because we exposed the Webdriver internal

src/Selenium2Driver.php
```diff
@@ -75,10 +81,54 @@ class Selenium2Driver extends CoreDriver
...
+            }
+        }
+    }
+
+    /**
+     * Sets the timeouts to apply to the webdriver session
+     *
+     * @param array $timeouts The session timeout settings: Array of {script, implicit, page} => time in milliseconds
+     */
+    public function setTimeouts($timeouts)
```
stof: shouldn't it be typehinted as array ?

src/Selenium2Driver.php
```diff
@@ -97,54 +147,20 @@ protected function setBrowserName($browserName = 'firefox')
...
-                    $chromeOptions[$capability] = $value;
-                }
-                $desiredCapabilities['chrome.'.$capability] = $value;
-            }
-
-            $desiredCapabilities['chromeOptions'] = $chromeOptions;
-
-            unset($desiredCapabilities['chrome']);
+        if ($this->started) {
+            throw new DriverException('Unable to set desiredCapabilities, the session has already started');
```
stof: I would keep that first

src/Selenium2Driver.php
```diff
@@ -163,23 +179,13 @@ public function getDesiredCapabilities()
     /**
      * Sets the WebDriver instance
      *
-     * @param WebDriver $webDriver An instance of the WebDriver class
+     * @param RemoteWebDriver $webDriver An instance of the WebDriver class
      */
-    public function setWebDriver(WebDriver $webDriver)
+    public function setWebDriver(RemoteWebDriver $webDriver)
```
stof: does this setter even make sense ? RemoteWebDriver is an equivalent of the session in the instaclick library. The class set there previously is the factory for the session (which is not an object which exists in the facebook library as they use a static method instead)

src/Selenium2Driver.php
```diff
@@ -260,7 +213,8 @@ protected static function charToOptions($char, $modifier = null)
      */
     protected function executeJsOnXpath($xpath, $script, $sync = true)
```
stof: As it is a new major version, I suggest switching to private too (we don't want to make those supported extension points)

src/Selenium2Driver.php
```diff
@@ -260,7 +213,8 @@ protected static function charToOptions($char, $modifier = null)
...
     {
-        return $this->executeJsOnElement($this->findElement($xpath), $script, $sync);
+        $element = $this->findElement($xpath);
```
stof: why changing this code ?

src/Selenium2Driver.php
```diff
@@ -297,46 +246,19 @@ private function executeJsOnElement(Element $element, $script, $sync = true)
...
-    private function applyTimeouts()
-    {
-        try {
-            foreach ($this->timeouts as $type => $param) {
-                $this->wdSession->timeouts($type, $param);
-            }
-        } catch (UnknownError $e) {
-            throw new DriverException('Error setting timeout: ' . $e->getMessage(), 0, $e);
-        }
+        $this->started = true;
```
stof: I suggest dropping that boolean entirely, relying on `null !== $this->webDriver` instead to check for being started

src/Selenium2Driver.php
```diff
@@ -434,75 +370,71 @@ public function switchToIFrame($name = null)
...
     }
 
     /**
      * {@inheritdoc}
      */
     public function getContent()
     {
-        return $this->wdSession->source();
+        $source = $this->webDriver->getPageSource();
+        return str_replace(array("\r", "\r\n", "\n"), \PHP_EOL, $source);
```
stof: I would rather return them with `"\n"` rather than `\PHP_EOL` if we normalize it

src/Selenium2Driver.php
```diff
@@ -560,58 +518,33 @@ public function getAttribute($xpath, $name)
...
-            break;
-        }
-    }
-}
-
-return value;
-JS;
-
-            return $this->executeJsOnElement($element, $script);
+            $element = new WebDriverRadios($element);
```
stof: I would not override the `$element` variable. `WebDriverRadios` is not the element, but a different helper.

src/Selenium2Driver.php
```diff
@@ -560,58 +518,33 @@ public function getAttribute($xpath, $name)
...
-for (var i = 0; i < node.options.length; i++) {
-    if (node.options[i].selected) {
-        value.push(node.options[i].value);
-    }
-}
-
-return value;
-JS;
+        if ('select' === $elementName) {
+            $element = new WebDriverSelect($element);
```
stof: same here about not reusing the variable name

stof: and there are more places below.

src/Selenium2Driver.php
```diff
@@ -758,170 +695,135 @@ public function selectOption($xpath, $value, $multiple = false)
...
     {
-        try {
-            // Move the mouse to the element as Selenium does not allow clicking on an element which is outside the viewport
-            $this->wdSession->moveto(array('element' => $element->getID()));
-        } catch (UnknownCommand $e) {
-            // If the Webdriver implementation does not support moveto (which is not part of the W3C WebDriver spec), proceed to the click
-        }
-
-        $element->click();
+        // Move the mouse to the element as Selenium does not allow clicking on an element which is outside the viewport
```
stof: this comment is not relevant anymore, as you removed the corresponding code

src/Selenium2Driver.php
```diff
@@ -758,170 +695,135 @@ public function selectOption($xpath, $value, $multiple = false)
...
+        $element = $this->findElement($xpath);
+        $this->webDriver->action()->contextClick($element)->perform();
     }
 
     /**
      * {@inheritdoc}
      */
     public function attachFile($xpath, $path)
     {
+        /** @var RemoteWebElement $element */
```
stof: this phpdoc should be useless. The return type should already be documented properly.

.travis.yml
```diff
@@ -1,43 +1,70 @@
...
+cache:
+  directories:
+    - $HOME/.composer/cache/files
+    - jar
 
 matrix:
-  fast_finish: true
   include:
-    - php: 7.0
-      env: WEBDRIVER=selenium-remote
```
stof: can we keep a test corresponding to that case ? It was important to test for people relying on Saucelabs to run their Selenium

src/Selenium2Driver.php
```diff
@@ -163,23 +179,13 @@ public function getDesiredCapabilities()
...
     {
         $this->webDriver = $webDriver;
     }
 
-    /**
-     * Gets the WebDriverSession instance
-     *
-     * @return \WebDriver\Session
-     */
-    public function getWebDriverSession()
```
stof: to allow bypassing the Mink abstraction for things which are not supported (accepting prompts for instance), we should expose `$this->webDriver` (which is the equivalent of what `\WebDriver\Session` is in the old library)

src/Selenium2Driver.php
```diff
@@ -758,170 +695,135 @@ public function selectOption($xpath, $value, $multiple = false)
...
      * {@inheritdoc}
      */
     public function focus($xpath)
     {
-        $this->trigger($xpath, 'focus');
+        $element = $this->findElement($xpath);
+        $action = $this->webDriver->action();
+
+        $action->moveToElement($element);
+        $action->click($element);
```
stof: clicking an element is not the same than focusing it. It triggers extra actions (for a link, it might navigate for instance)

src/Selenium2Driver.php
```diff
@@ -758,170 +695,135 @@ public function selectOption($xpath, $value, $multiple = false)
...
     }
 
     /**
      * {@inheritdoc}
      */
     public function keyDown($xpath, $char, $modifier = null)
     {
-        $options = self::charToOptions($char, $modifier);
-        $this->trigger($xpath, 'keydown', $options);
+        $this->sendKey($xpath, $char, $modifier);
```
stof: that's not the same API than previously. After calling keyDown, your implementation already released the key.

I think we should rather add a new `pressKey` method in Mink, and deprecate these 3 methods. And the new version of the driver would then implement only `pressKey`.
And we should implement the new method in 1.x too

The old implementation was not good anyway, as it was not actually pressing the key, but only triggering the JS actions (so the default behavior of pressing a key would not happen).

src/Selenium2Driver.php
```diff
@@ -934,7 +836,21 @@ public function executeScript($script)
             $script = '(' . $script . ')';
         }
 
-        $this->wdSession->execute(array('script' => $script, 'args' => array()));
+        $this->webDriver->executeScript($script);
+    }
+
+    public function executeAsyncScript($script)
```
stof: We don't have such method in the Mink API. Should we really add it ?

src/Selenium2Driver.php
```diff
@@ -946,50 +862,68 @@ public function evaluateScript($script)
...
-        return $this->wdSession->execute(array('script' => $script, 'args' => array()));
+        return $this->webDriver->executeScript($script);
     }
 
     /**
-     * {@inheritdoc}
+     * Waits some time or until JS condition turns true.
+     *
+     * @param int                                        $timeout   timeout in milliseconds
+     * @param WebDriverExpectedCondition|\Closure|string $condition JS condition / Closure
```
stof: I don't think it makes sense to change the documented signature compared to interface

src/Selenium2Driver.php
```diff
@@ -946,50 +862,68 @@ public function evaluateScript($script)
...
      * {@inheritdoc}
      */
     public function resizeWindow($width, $height, $name = null)
     {
-        $this->wdSession->window($name ? $name : 'current')->postSize(
-            array('width' => $width, 'height' => $height)
-        );
+        $dimension = new WebDriverDimension($width, $height);
+        if ($name) {
+            throw new \Exception('Named windows are not supported yet');
```
stof: should be an `UnsupportedDriverActionException` instead

src/Selenium2Driver.php
```diff
@@ -946,50 +862,68 @@ public function evaluateScript($script)
...
     }
 
     /**
      * {@inheritdoc}
      */
     public function maximizeWindow($name = null)
     {
-        $this->wdSession->window($name ? $name : 'current')->maximize();
+        if ($name) {
+            throw new \Exception('Named window is not supported');
```
stof: should be an `UnsupportedDriverActionException` instead

src/Selenium2Driver.php
```diff
@@ -999,216 +933,107 @@ public function maximizeWindow($name = null)
...
+     * @param WebDriverElement $element
+     * @param string           $xpath
+     * @param string           $type
+     * @param string           $action
      *
-     * @throws DriverException when the value cannot be found
+     * @throws DriverException
      */
-    private function selectRadioValue(Element $element, $value)
+    private function  ensureInputType(WebDriverElement $element, $xpath, $type, $action)
```
stof: you have an extra space here

src/Selenium2Driver.php
```diff
@@ -999,216 +933,107 @@ public function maximizeWindow($name = null)
...
-                $option->click();
-            }
-
-            return;
-        }
-
-        // Deselect all options before selecting the new one
-        $this->deselectAllOptions($element);
-        $option->click();
+        throw new \RuntimeException('Not yet supported');
```
stof: is this even called anywhere ? If no, remove the private method

src/Selenium2Driver.php
```diff
@@ -999,216 +933,107 @@ public function maximizeWindow($name = null)
...
+        }
 
-        $this->executeJsOnElement($element, $script);
-    }
+        return $modifier;
+}
 
     /**
-     * Ensures the element is a checkbox
+     * Decode char
```
stof: should be `Decodes`

src/Selenium2Driver.php
```diff
@@ -999,216 +933,107 @@ public function maximizeWindow($name = null)
...
-
-        // Deselect all options before selecting the new one
-        $this->deselectAllOptions($element);
-        $option->click();
+        throw new \RuntimeException('Not yet supported');
     }
 
     /**
-     * Deselects all options of a multiple select
+     * Prepend modifier
```
stof: should be `Prepends`

stof:
As that's a rewrite almost from scratch, should we keep it in a package named `selenium2-driver`, which is confusing as it is not about Selenium **2** anymore (but about W3C webdriver) ?

We may bring it as a new driver instead (and deprecate `selenium2-driver` in favor of the new one for people not needing to run tests on unmaintained PHP versions). This way, we could name it with a better name. What do you think ?

oleg-andreyev:
@stof I'm all for creating a new driver and deprecating this one, but as you mentioned in https://github.com/instaclick/php-webdriver/issues/84#issuecomment-452384343 facebook/php-webdriver is an abstraction for JsonWire (deprecated) and W3C, so I'd suggest naming new driver simply as `MinkWebDriver` (minkphp/webdriver)

.travis.yml
```diff
@@ -1,43 +1,70 @@
...
-      env: MINK_PHP_BIN=~/.phpenv/versions/5.6/bin/php
+    # XVFB with regular Chrome (not headless)
+    - php: 7.2
+      env:
+        - DRIVER_OPTIONS='{"args":["no-sandbox","window-size=1024,768"]}'
+        - START_XVFB=1
+    # build with lowest deps
+    - php: 7.2
+      env:
+        - DEPENDENCIES="--prefer-lowest"
```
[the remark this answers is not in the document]
oleg-andreyev: 👍

composer.json
```diff
@@ -18,29 +18,36 @@
...
-
     "autoload-dev": {
         "psr-4": {
             "Behat\\Mink\\Tests\\Driver\\": "tests"
         }
     },
-
     "extra": {
         "branch-alias": {
             "dev-master": "1.3.x-dev"
```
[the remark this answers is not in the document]
oleg-andreyev: 👍

phpunit.xml.dist
```diff
@@ -11,15 +14,26 @@
     <php>
         <var name="driver_config_factory" value="Behat\Mink\Tests\Driver\Selenium2Config::getInstance" />
 
-        <!--server name="WEB_FIXTURES_HOST" value="http://test.mink.dev" /-->
-        <!--server name="WEB_FIXTURES_BROWSER" value="firefox" /-->
+        <!-- Docker for Mac, see https://docs.docker.com/docker-for-mac/networking/#there-is-no-docker0-bridge-on-macos#i-want-to-connect-from-a-container-to-a-service-on-the-host -->
+        <!--<server name="WEB_FIXTURES_HOST" value="http://host.docker.internal:8002"/>-->
+        <server name="WEB_FIXTURES_HOST" value="http://localhost:8002"/>
```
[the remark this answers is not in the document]
oleg-andreyev: commented out and added comment where to find default value

tests/Selenium2Config.php
```diff
@@ -34,12 +71,31 @@ public function skipMessage($testCase, $test)
             return 'WebDriver does not support setting value in color inputs. See https://code.google.com/p/selenium/issues/detail?id=7650';
         }
 
+        $desiredCapabilities = $this->driver->getDesiredCapabilities();
+        $chromeOptions = $desiredCapabilities->getCapability(ChromeOptions::CAPABILITY);
+
+        /** @var ChromeOptions $capability */
```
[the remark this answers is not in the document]
oleg-andreyev: removed

.travis.yml
```diff
@@ -1,43 +1,70 @@
...
-  - vendor/bin/mink-test-server > /dev/null 2>&1 &
+before_script:
+  - mkdir ./logs
+  - if [[ "$BROWSER_NAME" = "chrome" && "$CHROMEDRIVER_VERSION" = "latest" ]]; then CHROMEDRIVER_VERSION=`curl -sS chromedriver.storage.googleapis.com/LATEST_RELEASE`; fi
+  - if [ "$BROWSER_NAME" = "chrome" ]; then mkdir chromedriver; wget -q -t 3 https://chromedriver.storage.googleapis.com/$CHROMEDRIVER_VERSION/chromedriver_linux64.zip; unzip chromedriver_linux64 -d chromedriver; fi
+  - if [ "$START_XVFB" = "1" ]; then sh -e /etc/init.d/xvfb start; fi;
+  - ./chromedriver/chromedriver --port=4444 --url-base=wd/hub --verbose &> ./logs/chromedriver.log &
+  - until $(echo | nc localhost 4444); do sleep 1; echo Waiting for ChromeDriver on port 4444...; done; echo "ChromeDriver started"
+  # https://bugs.php.net/bug.php?id=71443
+  - USE_ZEND_ALLOC=0 travis_retry php -d memory_limit=8G -S localhost:8002 -t ./vendor/mink/driver-testsuite/web-fixtures &> ./logs/mink-test-server.log &
```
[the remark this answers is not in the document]
oleg-andreyev: done https://github.com/minkphp/driver-testsuite/pull/28

README.md
```diff
@@ -49,6 +49,29 @@ $> curl -sS http://getcomposer.org/installer | php
...
+    docker run -p 4444:4444 selenium/standalone-firefox:2.53.1
+    ```
+    2. If you do not have Docker, but you have Java
+    ```bash
+    curl -L http://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.1.jar > selenium-server-standalone-2.53.1.jar
+    java -jar selenium-server-standalone-2.53.1.jar
+    ```
+2. Start WebServer by running
+    ``` bash
+    php -S localhost:8002 -t ./vendor/mink/driver-testsuite/web-fixtures
```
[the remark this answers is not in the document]
oleg-andreyev: done

src/Selenium2Driver.php
oleg-andreyev: done

src/Selenium2Driver.php
oleg-andreyev: done

src/Selenium2Driver.php
oleg-andreyev: removed this method

src/Selenium2Driver.php
oleg-andreyev: returned as `getWebDriver`

src/Selenium2Driver.php
oleg-andreyev: done

src/Selenium2Driver.php
oleg-andreyev: easier to debug and set breakpoint

src/Selenium2Driver.php
oleg-andreyev: agree

src/Selenium2Driver.php
oleg-andreyev: IIRC if you try to execute phpunit on Windows, it will fail on multi-line assert as example: `\Behat\Mink\Tests\Driver\Form\Html5Test::testHtml5FormButtonAttribute`

src/Selenium2Driver.php
oleg-andreyev: 👍

src/Selenium2Driver.php
oleg-andreyev: 👍

src/Selenium2Driver.php
oleg-andreyev: removed, but jfyi this logic is just implemented in `\Facebook\WebDriver\Remote\RemoteMouse`

src/Selenium2Driver.php
oleg-andreyev: updated

src/Selenium2Driver.php
oleg-andreyev: will investigate, good catch

src/Selenium2Driver.php
oleg-andreyev: done

src/Selenium2Driver.php
oleg-andreyev: done

src/Selenium2Driver.php
oleg-andreyev: removed

src/Selenium2Driver.php
oleg-andreyev: this method removed.

src/Selenium2Driver.php
oleg-andreyev: updated comment

src/Selenium2Driver.php
oleg-andreyev: returned `inheritdoc`, but the method remains untouched (hidden functionality)

src/Selenium2Driver.php
oleg-andreyev: fixed

src/Selenium2Driver.php
stof: Well, I consider it easier to write tests when I don't need to change the newline chars in the test depending on the OS.

One reason it might fail for you is if your git is configured to convert the code to CRLF on checkout. But I don't consider that a good setup. The unmodified source code must be working.

src/Selenium2Driver.php
stof: but we don't have that API in Mink. We should not introduce it here.

oleg-andreyev:
@stof build is again green. I'll start working on deprecating keyUp/keyPress/keyDown in favor of pressKey as per discussion https://github.com/minkphp/MinkSelenium2Driver/pull/304#discussion_r246046293

stof:
@oleg-andreyev please do that in separate PRs, so that it can be applied to the existing driver too.

oleg-andreyev:
Meanwhile can you create “webdriver” repository under minkphp for new driver, thanks

src/Selenium2Driver.php
oleg-andreyev: https://github.com/minkphp/Mink/pull/772

oleg-andreyev:
I'm moving my effort to https://github.com/oleg-andreyev/MinkFacebookWebDriver

PR #309, opened 2019-06-21
Keep the chromeOptions array from configuration (w3c options)

macintoshplus:
In relation with the issue #293, this change allow keep in the array chrome capabilites all options set into the behat configuration (key `default.extensions.Behat\MinkExtension.sessions.selenium2.capabilities.extra_capabilities.chromeOptions`)

Now if you set this option:
```
  extensions:
    Behat\MinkExtension:
      selenium2:
         capabilities: { "browser": "chrome", "browserName": "chrome", "extra_capabilities": { "chromeOptions": { "w3c": false } } }
```

The option will be send for the session option with ChromeDriver 75.0.3770.8+

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -131,7 +131,7 @@ class Selenium2Driver extends CoreDriver
         // See https://sites.google.com/a/chromium.org/chromedriver/capabilities
         if (isset($desiredCapabilities['chrome'])) {
 
-            $chromeOptions = (isset($desiredCapabilities['chromeOptions']) && is_array($desiredCapabilities['chromeOptions']))? $desiredCapabilities['chromeOptions']:array();
+            $chromeOptions = (isset($desiredCapabilities['goog:chromeOptions']) && is_array($desiredCapabilities['goog:chromeOptions']))? $desiredCapabilities['goog:chromeOptions']:array();
 
             foreach ($desiredCapabilities['chrome'] as $capability => $value) {
                 if ($capability == 'switches') {
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -131,7 +131,7 @@ class Selenium2Driver extends CoreDriver
         // See https://sites.google.com/a/chromium.org/chromedriver/capabilities
         if (isset($desiredCapabilities['chrome'])) {
 
-            $chromeOptions = array();
+            $chromeOptions = (isset($desiredCapabilities['chromeOptions']) && is_array($desiredCapabilities['chromeOptions']))? $desiredCapabilities['chromeOptions']:array();
 
             foreach ($desiredCapabilities['chrome'] as $capability => $value) {
                 if ($capability == 'switches') {


src/Selenium2Driver.php
```diff
@@ -131,7 +131,7 @@ public function setDesiredCapabilities($desiredCapabilities = null)
         // See https://sites.google.com/a/chromium.org/chromedriver/capabilities
         if (isset($desiredCapabilities['chrome'])) {
 
-            $chromeOptions = array();
+            $chromeOptions = (isset($desiredCapabilities['chromeOptions']) && is_array($desiredCapabilities['chromeOptions']))? $desiredCapabilities['chromeOptions']:array();
 
             foreach ($desiredCapabilities['chrome'] as $capability => $value) {
```
aik099: The `w3c` option should go into `$desiredCapabilities['chrome']` array to work correctly.

Are you sure Behat (through MinkExtension) have no way of populating that array?

Here is relevant code fragment, that builds that array for the driver on Behat side:

https://github.com/Behat/MinkExtension/blob/master/src/Behat/MinkExtension/ServiceContainer/Driver/Selenium2Factory.php#L51-L86

src/Selenium2Driver.php
```diff
@@ -131,7 +131,7 @@ public function setDesiredCapabilities($desiredCapabilities = null)
         // See https://sites.google.com/a/chromium.org/chromedriver/capabilities
         if (isset($desiredCapabilities['chrome'])) {
 
-            $chromeOptions = array();
+            $chromeOptions = (isset($desiredCapabilities['chromeOptions']) && is_array($desiredCapabilities['chromeOptions']))? $desiredCapabilities['chromeOptions']:array();
```
aik099: That adds another layer of complexity, because there is already `$desiredCapabilities['chrome']` array with all chrome-specific options.

aik099:
Does this configuration without your PR solve the issue:

```yml
  extensions:
    Behat\MinkExtension:
      selenium2:
         capabilities: { "browser": "chrome", "browserName": "chrome", "chrome": { "w3c": false } }
```

?

macintoshplus:
> Does this configuration without your PR solve the issue:
> 
> ```yaml
>   extensions:
>     Behat\MinkExtension:
>       selenium2:
>          capabilities: { "browser": "chrome", "browserName": "chrome", "chrome": { "w3c": false } }
> ```
> 
> ?

Nope, I have this exception:
```
[Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]                   
  Unrecognized option "w3c" under "testwork.mink.sessions.default.selenium2.capabilities.chrome"
```

src/Selenium2Driver.php
macintoshplus: The code fragment, show how the content of the array `extra_capabilities` defined into the Behat configuration will be add into the array fo `$desiredCapabilities` variable.

But the line `$chromeOptions = array();` initilize an new array without check if the current configuration have already this array.
And the line `$desiredCapabilities['chromeOptions'] = $chromeOptions;` set an new value without merge the content.

src/Selenium2Driver.php
macintoshplus: Another way is to use the `array_merge` at the line https://github.com/minkphp/MinkSelenium2Driver/blob/master/src/Selenium2Driver.php#L145
for not erase the content.

The option `w3c` must be set into the `chromeOptions` for disable the W3C WebDriver dialect for ChromeDriver.

macintoshplus:
The goal of the configuration key `Behat\MinkExtension.sessions.selenium2.capabilities.extra_capabilities` is for add extra options without need to change Mink or MinkExtension. It's wrong ?

src/Selenium2Driver.php
macintoshplus: @aik099 are my answers satisfactory?

macintoshplus:
@aik099 are my answers satisfactory?

aik099:
> The goal of the configuration key `Behat\MinkExtension.sessions.selenium2.capabilities.extra_capabilities` is for add extra options without need to change Mink or MinkExtension. It's wrong ?

I guess so. I'm not using Behat myself though. Only Mink.

> @aik099 are my answers satisfactory?

That does shed some light on circumstances, that led you to writing that exact code. We'll wait for @stof opinion on this one.

macintoshplus:
ping @stof

stof:
should be updated to use the `goog:chromeOptions`, which is the name supported in recent versions of ChromeDriver (`chromeOptions` was deprecated and has been dropped)

aik099:
According to https://github.com/Codeception/Codeception/pull/5349/files actual rename happened in Selenium 3.8.

macintoshplus:
Change done.

src/Selenium2Driver.php
```diff
@@ -142,7 +142,7 @@ public function setDesiredCapabilities($desiredCapabilities = null)
                 $desiredCapabilities['chrome.'.$capability] = $value;
             }
 
-            $desiredCapabilities['chromeOptions'] = $chromeOptions;
+            $desiredCapabilities['goog:chromeOptions'] = $chromeOptions;
```
aik099: Please rebase, because this change was already made in #312.

src/Selenium2Driver.php
```diff
@@ -131,7 +131,7 @@ public function setDesiredCapabilities($desiredCapabilities = null)
         // See https://sites.google.com/a/chromium.org/chromedriver/capabilities
         if (isset($desiredCapabilities['chrome'])) {
 
-            $chromeOptions = array();
+            $chromeOptions = (isset($desiredCapabilities['goog:chromeOptions']) && is_array($desiredCapabilities['goog:chromeOptions']))? $desiredCapabilities['goog:chromeOptions']:array();
```
aik099: I'm not sure if we should rename keys that come from `.behat.yml`. Maybe `:` in YAML keys isn't even supported.

macintoshplus: Yes, this change introduce a backward compatibility break. It is possible to keep the older option name and change the option name here only after the process.
What doing ?

src/Selenium2Driver.php
macintoshplus: Rebase done !

src/Selenium2Driver.php
aik099: Since before your PR options weren't passed at all, then no BC break really :) Looks good.

macintoshplus:
Thanks @aik099.

The next version of driver will be compatible only with Chrome 74+ ?

aik099:
> The next version of driver will be compatible only with Chrome 74+ ?

@macintoshplus , because of #312 it will. I'll merge your PR then. Thanks.

PR #312, opened 2019-08-29
Use goog:chromeOptions instead of chromeOptions

goetas:
See https://github.com/SeleniumHQ/docker-selenium/issues/674


goetas:
Thanks!

PR #320, opened 2020-07-04
Fix zip file creation on PHP 8

alexpott:
Using MinkSelenium2Driver on PHP 8 results in deprecation warnings due to using ZipArchive with an empty file - see https://github.com/php/php-src/blob/master/UPGRADING#L666

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -1189,7 +1189,10 @@ JS;
         }
 
         // Selenium only accepts uploads that are compressed as a Zip archive.
-        $tempFilename = tempnam('', 'WebDriverZip');
+        // PHP 8 has deprecated using ZipArchive with an empty file. Deleting
+        // the file could lead to problems during concurrent testing so add .zip
+        // so the file does not exist.
+        $tempFilename = tempnam('', 'WebDriverZip') . '.zip';
 
         $archive = new \ZipArchive();
         $result = $archive->open($tempFilename, \ZipArchive::CREATE);
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -1189,13 +1189,10 @@ JS;
         }
 
         // Selenium only accepts uploads that are compressed as a Zip archive.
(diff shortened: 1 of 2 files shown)


src/Selenium2Driver.php
```diff
@@ -1189,7 +1189,10 @@ private function uploadFile($path)
         }
 
         // Selenium only accepts uploads that are compressed as a Zip archive.
-        $tempFilename = tempnam('', 'WebDriverZip');
+        // PHP 8 has deprecated using ZipArchive with an empty file. Deleting
+        // the file could lead to problems during concurrent testing so add .zip
+        // so the file does not exist.
+        $tempFilename = tempnam('', 'WebDriverZip') . '.zip';
```
oleg-andreyev, line 1195: To reduce I/O, I'd replace it with `sys_get_temp_dir` and `uniqid`

alexpott: Do we trust uniqid() to generate a unique ID across concurrant processes that call it at exactly the same time - https://www.php.net/manual/en/function.uniqid.php . The advantage of tempnam is that we know the name is unique for the given filesystem.

oleg-andreyev: You can set second argument “more_entropy” it’s more likely to be unique. 

@stof, @aik099 what do you think?

aik099: I see these ways of solving the problem:

1. use CSPRNG (like `random_bytes` or reading from `/dev/urandom`) to ensure unique filename without needing to create a lock file using `tempnam`
2. try using `\ZipArchive::OVERWRITE` instead of `\ZipArchive::CREATE` and see if that works with an empty file

With the current solution nobody guarantees, that file with `$tempFilename` name actually exists already.

alexpott: \ZipArchive::OVERWRITE works great. Nice.

aik099 (approved):

aik099:
Merging, thanks @alexpott .

Issue #324, opened 2020-10-05
Changes in master have broken Drupal's test suite


alexpott:
#322 and #323 have broken Drupal's Javascript testing. If I use the master branch of the driver unless I revert these two changes tests that used to pass now fail.

alexpott:
#322 is breaking because it looks like we're not quite waiting enough for things to be interactive.

alexpott:
#323 is breaking because we are passing the following to condition.
```
    $condition = <<<JS
      (function() {
        function isAjaxing(instance) {
          return instance && instance.ajaxing === true;
        }
        return (
          // Assert no AJAX request is running (via jQuery or Drupal) and no
          // animation is running.
          (typeof jQuery === 'undefined' || (jQuery.active === 0 && jQuery(':animated').length === 0)) &&
          (typeof Drupal === 'undefined' || typeof Drupal.ajax === 'undefined' || !Drupal.ajax.instances.some(isAjaxing))
        );
      }());
JS;
```
Which used to work prior to #323

alexpott:
I think it could be argued that both fixes could be done in Drupal, however say the next release is 1.41 and that contains the PHP 8 fix #320 I'm not sure a bugfix release should contain them.

alexpott:
Here's what it would take to make Drupal's tests pass on master - https://www.drupal.org/project/drupal/issues/3174928

stof:
I think adding a `rtrim($condition, ';')` in `wait()` might fix that case.

stof:
@alexpott do you want to send a PR for this fix ?

Note that I still suggest to land the Drupal patch though, as the argument is meant to be an expression and so the `;` should not be included in it.

stof:
Closing this as the BC break of #323 has been solved, and the timeout issue has been solved on Drupal (to stop implicitly relying on the fact that `->waitFor()` was waiting long by default and so replacing an explicit wait necessary before the first check in case the JS needed time to react to the button click)

PR #325, opened 2020-10-05
Trim trailing semi-colons for BC

alexpott:
Fixes part of #324

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -972,8 +972,6 @@ JS;
      */
     public function wait($timeout, $condition)
     {
-        // Remove trailing semi-colons for maximum compatibility with previous
-        // releases.
         $script = 'return (' . rtrim($condition, ';') . ');';
         $start = microtime(true);
         $end = $start + $timeout / 1000.0;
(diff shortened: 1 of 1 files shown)


mvorisek:
I am the author of the original PR and I think this BC fix is ok.

I would probably remove the comment, as it is quite obvious what it does and blame/history can be used for more info.

src/Selenium2Driver.php
```diff
@@ -972,7 +972,7 @@ public function evaluateScript($script)
      */
     public function wait($timeout, $condition)
     {
-        $script = 'return (' . $condition . ');';
+        $script = 'return (' . rtrim($condition, ';') . ');';
```
mvorisek, line 975:
```suggestion
        $script = 'return (' . rtrim(rtrim($condition), ';') . ');';
```

stof (approved):

src/Selenium2Driver.php
mvorisek, line 975:
```suggestion
        $script = 'return (' . rtrim($condition, " \t\n\r;") . ');';
```

PR #326, opened 2020-10-06
Only wait when necessary

alexpott:
In \Behat\Mink\Driver\Selenium2Driver::wait we always wait even if the condition returns true. This results in unnecessary waiting.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -978,8 +978,11 @@ JS;
 
         do {
             $result = $this->wdSession->execute(array('script' => $script, 'args' => array()));
-            !$result && usleep(10000);
-        } while (!$result && microtime(true) < $end);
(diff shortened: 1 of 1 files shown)


src/Selenium2Driver.php
```diff
@@ -978,8 +978,8 @@ public function wait($timeout, $condition)
 
         do {
             $result = $this->wdSession->execute(array('script' => $script, 'args' => array()));
-            usleep(10000);
-        } while (microtime(true) < $end && !$result);
+            !$result && usleep(10000);
```
stof, line 981: I'd rather write it like that:

```php
do {
  $result = // ...

  if ($result) {
    break;
  }
while (microtime(true) < $end):
```

aik099:
Merging, thanks @alexpott .

PR #327, opened 2021-01-12
Fix getText() to work with elements outside viewport

mvorisek:
Original impl. using `$element->text()` is defined as "Returns the visible text for the element." [1].

This PR fixes the implementation to make`getText()` working with elements outside viewport. The new impl. is inspired by `getHtml()` method defined a few lines below. Thus this fix should be consistent and there should be no downside.

See https://github.com/atk4/ui/pull/1538 - when small window (like 1280x720) is given, the tests fail with the original implementation. With this PR, the issue is fixed.

[1] https://github.com/instaclick/php-webdriver/blob/v1.0.8/lib/WebDriver/Element.php#L33

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -539,8 +539,7 @@ class Selenium2Driver extends CoreDriver
      */
     public function getText($xpath)
     {
-        $node = $this->findElement($xpath);
-        $text = $node->text();
+        $text = $this->executeJsOnXpath($xpath, 'return {{ELEMENT}}.innerText;');
         $text = (string) str_replace(array("\r", "\r\n", "\n"), ' ', $text);
 
         return $text;


oleg-andreyev:
It's an expected behavior https://www.w3.org/TR/webdriver/#get-element-text

mvorisek:
> 
> 
> It's an expected behavior https://www.w3.org/TR/webdriver/#get-element-text
> 
> I don't think it's a responsibility of this package. You just need to make sure that text is in a viewport (same as your user)

For W3C driver conformant yes, but not for general element itself. As discussed in the PR desc, `getHtml()` always works and `getText()` should too - this PR fixes it.

mvorisek:
@stof @aik099 can you please merge this PR and do a bugfix release? Thanks.

mvorisek:
@stof @aik099 please review

aik099:
The Selenium Server (used by this driver) only operates page elements, that the user can see. For that reason, there can't be a Mink/Behat test, that tests parts of the page, invisible to the user.

Because of the above fact, Selenium Server API doesn't have a method to retrieve HTML of an element (user can't see it on the page anywhere) and we have to develop a JavaScript-based solution, which doesn't respect element visibility as you've noticed.

There are no tests for `getText`/`getHTML` asserting that they're accessing visible parts of the page because only the Selenium driver is capable of that.

I'm not certain what direction we should take with this:

1. make `getText` operate on invisible elements (as `getHTML` does)
2. change `getHTML` to also operate on only visible elements

// @stof 

Related PR: #328

mvorisek:
> I'm not certain what direction we should take with this:
> 
>     1. make `getText` operate on invisible elements (as `getHTML` does)
> 
>     2. change `getHTML` to also operate on only visible elements
> 

@aik099 Thank you for getting into it. I strongly belive we should go with the 1st option because:

a) some browsers like Firefox operate on invisible elements, some browsers like Chrome does not (and only in some modes, like `--headless`). With this PR, the behaviour for the `getText` is consistent.
b) other methods like `getHTML` already operate on invisible elements, thus the behaviour is then consistent across methods.

mvorisek:
@stof can you please review this PR

mvorisek:
@stof, @aik099 can we merge this PR please?

mvorisek:
Hi @stof, @aik099, guys, can you please merge this PR to allow easier/more consistent Behat testing?

Thanks in advance!

aik099:
> See atk4/ui#1538 - when small window (like 1280x720) is given, the tests fail with the original implementation. With this PR, the issue is fixed.

@mvorisek , the case you're facing (website doesn't fit in a small browser window and therefore it looks like the user isn't seeing the text) is expected behavior.

How you can hope to pass the test when text is invisible to the user?

P.S.
After thinking more about your proposition I still think, that operations outside of the viewport (which was made via JS just to conform with the DriverInterface really) isn't something, that this driver should ever do. I hope you'll understand.

mvorisek:
@aik099 I think of it this way - DOM has no viewport, if a DOM element is queried for a text with `getText`, a text of the element should be returned no matter if presented in a viewport or not. Secondly, similar functon `getHTML` does it already.

Chekote:
@mvorisek I think you're not using the right tool for the job. Selenium is for user testing. You should not be interacting with or making assertions against anything that the user cannot interact with or see.

You'll notice this being enforced as the Selenium server has evolved. In older versions, you used to be able to click elements that were covered by other elements, but with newer versions, you will now get an exception. If a user cannot do it, then Selenium won't allow you to do it either.

aik099:
Closing then.

PR #328, opened 2021-01-13
Fix NL replace (do not replace \r\n with two spaces)

mvorisek:


src/Selenium2Driver.php
```diff
@@ -541,7 +541,7 @@ public function getText($xpath)
     {
         $node = $this->findElement($xpath);
         $text = $node->text();
-        $text = (string) str_replace(array("\r", "\r\n", "\n"), ' ', $text);
+        $text = str_replace(array("\r\n", "\r", "\n"), ' ', $text);
```
mvorisek, line 544:
```suggestion
        $text = trim(preg_replace('~\s+~s', ' ', $text));
```

mvorisek:
@stof @aik099 please review

aik099:
**Problem uncovered by this PR:** we're not testing the normalization logic of the `getText` method, because:

* tests haven't failed after the change in this PR
* each of drives does it's own normalization (Zombie replaces any trailing whitespace with a single space; BrowserKit replaces `\n` with a single space and then replaces all trailing spaces with a single space; Selenium2 replace any line break (after PR except `\r\n` with a single space)

Proposed next steps:

1. decide what we can test (not sure if `\r` could properly be tested in headless drivers)
2. add a new test and restart all driver repo builds to see which drivers actually comply
3. rebase this PR to see how the Selenium2 driver is doing

src/Selenium2Driver.php
```diff
@@ -541,7 +541,7 @@ public function getText($xpath)
     {
         $node = $this->findElement($xpath);
         $text = $node->text();
-        $text = (string) str_replace(array("\r", "\r\n", "\n"), ' ', $text);
```
aik099, line 544: Won't swapping 2nd and 1st array element solve the issue as well?

aik099 (changes requested):

src/Selenium2Driver.php
mvorisek: `\r\n` to 2 spaces yes, but other whitespaces normalization not (like if some browser keeps tabs etc.). Plus this PR also trim the ends.

mvorisek:
@stof can you please review?

mvorisek:
I have removed https://github.com/minkphp/MinkSelenium2Driver/commit/4391f590d3b0fe0adc7033edc45900df0a61b1c3 commit to not trim and keep this PR to bare minimum of the described problem.

aik099:
@mvorisek , now I like the change.

@stof , what about testing considering that each driver is doing it differently?

aik099 (approved):

PR #329, opened 2021-01-27
avoid passing null to strtolower

cdaguerre:
Calling `$element->attribute('type')` on an `<option>` element returns `null`, passing `null` to `strtolower` triggers a PHP error.
Frankly, I don't know why this started to fail all of the sudden in our test suite. I'm guessing it's due to a different selenium version, but this fixes it. (`$elementType` is not used in the case of a `<select>` element).

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -579,7 +579,7 @@ class Selenium2Driver extends CoreDriver
     {
         $element = $this->findElement($xpath);
         $elementName = strtolower($element->name());
-        $elementType = strtolower($element->attribute('type'));
+        $elementType = strtolower($element->attribute('type') ?: '');
 
         // Getting the value of a checkbox returns its value if selected.
         if ('input' === $elementName && 'checkbox' === $elementType) {


aik099:
@cdaguerre , I've tried doing `strtolower(null)` on all PHP versions (see https://3v4l.org/X0WM5) and it doesn't trigger a Fatal Error.

Could you please share the error message you're getting with a backtrace and PHP version/OS you're using?

mvorisek:
Yes, this should be closed as strict types are not defined for that changed file.

stof:
@aik099 it will trigger a deprecation warning on 8.1

aik099:
Then this needs to be merged.

src/Selenium2Driver.php
```diff
@@ -579,7 +579,7 @@ public function getValue($xpath)
     {
         $element = $this->findElement($xpath);
         $elementName = strtolower($element->name());
-        $elementType = strtolower($element->attribute('type'));
+        $elementType = strtolower($element->attribute('type') ?: '');
```
alexpott, line 582: As we're dealing with nulls I think doing ```?? ''``` is probably better

alexpott: Lol no - the minimum PHP in compose.json is 5.4!

alexpott:
There are 4 instances of `strtolower($element->attribute('type'))` in the codebase - I guess they all need this change.

aik099:
> There are 4 instances of `strtolower($element->attribute('type'))` in the codebase - I guess they all need this change.

@alexpott , please change them all as well.

src/Selenium2Driver.php
aik099: Your code already deals with NULL, so no changes needed to this line I guess.

alexpott:
> > There are 4 instances of `strtolower($element->attribute('type'))` in the codebase - I guess they all need this change.
> 
> @alexpott , please change them all as well.

Done in https://github.com/minkphp/MinkSelenium2Driver/pull/334

aik099:
Closed via #334.

Issue #279, opened 2018-02-05
Queries re use of Syn library


andrewnicols:
#233 claimed to update to version 0.0.3 of the Syn library, however it actually just made unknown changes to 0.0.2.

I really think that we need to update to the latest version, and I think that we can do so relatively easily.

The Syn library has seen several changes between our custom version and 0.0.3, including a number of bug fixes (I've contributed to some of them).

The breaking changes that I'm aware of boil down to:
* renamed from `Syn` to `syn`, and
* changed argument order for all calls from `'eventname', {options}, {{ELEMENT}}` to `{{ELEMENT}}, 'eventname', {optionalOptions})`

Since they have made both of these changes, we can add a b/c layer which essentially does:
```
window.Syn = {
    trigger: function(eventName, options, element) {
        console.log("Deprecated. Update to window.syn and swap your arguments");
        return window.syn.trigger(element, eventName, options);
    }
};
```

andrewnicols:
Incidentally, I'd support updating to the latest version of Syn.

aik099:
1. I think `Syn` is internal stuff of the driver and BC layer isn't necessary. What you think @stof ?
2. @andrewnicols , if all tests, that use Syn currently pass you can safely send PR with updated Syn version.

stof:
@andrewnicols I think we can still update Syn in a minor version of the driver and document the change. I would not consider extending the driver to use Syn to be an intended usage of the driver. It should have been internal.

Regarding the BC layer, I'm not sure this logging message would be visible in a place where users are seeing it, so it would be worse IMO.

andrewnicols:
I suspected as much, but I hadn't had a chance to confirm. Thanks for the info.

I'll try and put the pull request together.

PR #333, opened 2021-05-12
Update Syn to 0.15.0

zrpnr:
The Syn library has been under active development, and the current release is 0.15.0. MinkSelenium2Driver uses 0.0.3 currently.
 In 0.15.0 there was a fix for the deprecated [which](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which) property.
https://github.com/bitovi/syn/issues/180

This affects any sites that are running tests with `keyPress` events if, for example- they use the latest jQuery 3.6.0 that removed a shim for this property.
https://github.com/jquery/jquery/pull/4765

There was discussion about updating Syn in:

- https://github.com/minkphp/MinkSelenium2Driver/issues/279
- https://github.com/minkphp/MinkSelenium2Driver/issues/20

The minified js in this PR comes from the `v0.15.0` tag of https://github.com/bitovi/syn
I added 
```
options: {
  minify: true
},
```
to the export command in `build.js` and ran `npm run build` to produce `dist/global/syn.js`

In `Selenium2Driver.php` updated `Syn` to `syn` and updated the `trigger` function since the order of arguments has changed.


aik099 (changes requested):
@zrpnr , please make sure tests are passing. Right now they're not.

mvorisek:
@aik099 #358 was merged, let's close this PR

aik099:
Closing in favor of #358 .

PR #334, opened 2021-07-19
Fix PHP 8.1 deprecation

alexpott:
Based on work in https://github.com/minkphp/MinkSelenium2Driver/pull/329

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -579,7 +579,7 @@ class Selenium2Driver extends CoreDriver
     {
         $element = $this->findElement($xpath);
         $elementName = strtolower($element->name());
-        $elementType = strtolower($element->attribute('type'));
+        $elementType = strtolower($element->attribute('type') ?: '');
 
         // Getting the value of a checkbox returns its value if selected.
         if ('input' === $elementName && 'checkbox' === $elementType) {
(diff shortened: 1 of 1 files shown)


src/Selenium2Driver.php
```diff
@@ -579,7 +579,7 @@ public function getValue($xpath)
     {
         $element = $this->findElement($xpath);
         $elementName = strtolower($element->name());
-        $elementType = strtolower($element->attribute('type'));
+        $elementType = strtolower($element->attribute('type') ?: '');
```
mvorisek, line 582:
```suggestion
        $elementType = strtolower($element->attribute('type') ?? '');
```

alexpott: @mvorisek yeah I thought this too but the minimum PHP version is 5.4 :)

alexpott: I thought about the dangers of using ?: - ie. something unexpected evaluating to something FALSY - i.e `'0'` but this is never going to be valid value for the type attribute and in most cases this would have no effect even if possible because of the comparisons we make later in the code.

alexpott: Here's the list of permitted types - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types - none of them are going to evaluate to anything FALSY in PHP land.

mvorisek: I belive raising minimal PHP version to 7.x now is ok, see https://3v4l.org/sZM1h

stof: based on https://packagist.org/packages/behat/mink-selenium2-driver/php-stats, we could probably drop support for PHP 5.x in a new minor version. But this still deserves being discussed in a dedicated issue rather than in the comments of a PR fixing PHP 8.1 issues (especially when it could be achieved without using `??`

aik099:
Merging, thanks @alexpott .

PR #336, opened 2021-08-24
Fix StaleElementException in setValue when input has gone away

Chekote:
## Problem
Some websites using a SPA approach have implemented behavior where when a correct value is entered into an input, it is immediately recognized and the SPA updates instantly, causing the input to be detached from the DOM.

e.g. a 2FA input field that immediately recognizes the correct value as it is typed, and causes the SPA to update the page instantly if it is correct.

When this happens, the input is set correctly via the `$element->postValue()` call from `setValue()`, but when it then proceeds to call `$this->executeJsOnElement()`, a StaleElementException is thrown because the input is no longer attached to the DOM.

## Fix
Catch and swallow the exception.

## Rationale
It is not possible to execute JS on an element that is detached from the DOM, so it is impossible to attempt to unfocus it at this point (at least by executing JS on it).

The goal is to unfocus the element, which if it is detached from the DOM, it is clearly unfocused at this point.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -14,6 +14,7 @@ use Behat\Mink\Exception\DriverException;
 use Behat\Mink\Selector\Xpath\Escaper;
 use WebDriver\Element;
 use WebDriver\Exception\NoSuchElement;
+use WebDriver\Exception\StaleElementReference;
 use WebDriver\Exception\UnknownCommand;
 use WebDriver\Exception\UnknownError;
 use WebDriver\Exception;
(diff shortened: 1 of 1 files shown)


Chekote:
Seems like your Travis CI is broken? It didn't run on the PR.

Chekote:
@aik099 sorry to bug you, but I wanted to make sure you saw this PR since your CI seems to be broken.

src/Selenium2Driver.php
```diff
@@ -707,7 +708,24 @@ public function setValue($xpath, $value)
...
+        // functionality whereby entering a specific value could cause the field
+        // to be detached from the DOM.
+        //
+        // e.g. a 2FA input field that immediately recognizes the correct value
+        // as it is typed, and causes the SPA to update the page instantly.
+        //
+        // If that happens, we will get a StaleElementReference exception when
+        // we try to execute JS on it. As such, we catch that condition here and
+        // ignore it since an element is obviously no longer focused if it is
+        // no longer attached to the page document.
```
aik099, line 722: Please replace with:

```php
// Cover case, when an element was removed from DOM after its value was
// changed (e.g. by a JavaScript of a SPA) and therefore can't be focused.
```

---

People always can use git annotate to locate associated PR, which explains all in much greater detail.

src/Selenium2Driver.php
```diff
@@ -707,7 +708,24 @@ public function setValue($xpath, $value)
...
+        try {
+            $this->executeJsOnElement($element, $script);
+        } catch (StaleElementReference $e) {
+            // Do nothing. The element is no longer attached to the page document
+            // and therefore is no longer focused (nor can it be blurred).
```
aik099, line 727: Please replace with:
```php
// Do nothing because an element was already removed and therefore
// blurring is not needed.
```

aik099 (changes requested):

aik099:
Nice bug report by the way. Very detailed. Thank you @Chekote .

aik099 (approved):

aik099:
Merging, thanks @Chekote .

PR #337, opened 2021-09-08
Remove incorrect comment

Chekote:
This comment is stating that the code is adding a tab key, but it is not.


aik099:
Actual TAB triggering code was removed in #244.

aik099:
Merging, thanks @Chekote .

PR #339, opened 2021-10-05
Move to GitHub actions

alexpott:
This is necessary to test on PHP 8.1.

There is a pre-requisite PR for minkphp/driver-testsuite - see  https://github.com/minkphp/driver-testsuite/pull/46

I couldn't get the tests to work on PHP 7.1 for some unknown reason but given tests work on every other version of PHP fromo 5.4 to 8.1 - see https://github.com/alexpott/MinkSelenium2Driver/actions/runs/1305769251


.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: shivammathur/setup-php@v2
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php }}"
+          tools: composer
+
+      - name: Configure for PHP >= 8.0
+        if: "${{ matrix.php >= '8.0' }}"
+        run: |
+          composer config platform.php 7.4.99
```
stof, line 42: Why do we need this ?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          composer update --no-interaction --prefer-dist
+
+      - name: Setup Mink test server
+        run: |
+          mkdir ./logs
+          ./vendor/bin/mink-test-server &> ./logs/mink-test-server.log &
+
+      - name: Start Selenium
+        shell: bash
```
stof, line 59: useless as that's already the default shell you configured

.github/workflows/tests.yml
alexpott: It was in the travis file. Let's remove it.

stof (approved):

aik099:
@alexpott , please restore test suite coverage collection reporting.

// @stof

PR #341, opened 2021-10-10
Reinstate code coverage

alexpott:
@aik099 requested this. @stof asked me to remove it from the original MR but thinking about it we should only remove it in the MR the replaces it with psalm / phpstan

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,12 +27,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
         with:
-          fetch-depth: 1
+          fetch-depth: 2
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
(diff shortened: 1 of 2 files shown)


andypost:
Somehow Scrutinizer is stuck reporting result back to PR

alexpott:
I got scruntinizer config from a previous build where it worked and dumped that in a .scutinizer.yml file. My guess is that the global config or project config for this project on scutinizer's site is somehow messed up. Anyway the upshot is that it is fixed.

aik099:
@stof , any idea why we're running "PHPUnit with Coverage" for every PHP version instead of just running PHPUnit for every PHP version and for a single version run "PHPUnit with Coverage"?

It would only make sense to me when code has lots of PHP version-specific IFs, but we don't have this in Mink.

stof:
Closing in favor of #344

PR #342, opened 2021-10-12
Fix NavigationTest::testLinks on PHP 7.1

alexpott:
This fixes a segfault that occurs on PHP 7.1 when date.timezone is not configured.

In testing this has not proved enough because another segfault occurs in \Behat\Mink\Tests\Driver\Form\GeneralTest::testBasicForm() after running NavigationTest::testLinks(). What's really odd is if you run them the other way around it works just fine.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,8 @@ jobs:
           coverage: "none"
           php-version: "${{ matrix.php }}"
           tools: composer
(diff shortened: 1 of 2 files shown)


alexpott:
https://github.com/marketplace/actions/setup-php-action doesn't support patch versions so there's no new info from the last commit.

stof:
@alexpott I suggest giving up on this one, as you don't find a way to avoid the segfault on PHP 7.1

PR #344, opened 2021-12-10
Add code coverage with codecov

stof:


stof:
I'm merging this even though the CI fails for the new test added in https://github.com/minkphp/driver-testsuite/pull/15. This will be handled separately.

Issue #347, opened 2022-01-19
BasicAuthTest is failing in CI


stof:
`BasicAuthTest::testBasicAuthInUrl` is failing.

This might be an issue in the driver-testsuite regarding proper isolation of basic auth tests, as browsers maintain state of basic auth.

PR #348, opened 2022-01-19
Drop support for PHP <7.2

stof:


stof:
I'm merging this. The CI is red, but there is only one failing test which is already failing in master (see #347)

PR #352, opened 2022-04-15
Convert element to NodeElement from execute result

mvorisek:
no BC break (`\Behat\Mink\Driver\DriverInterface` interface cannot be updated in `behat/mink` v1.x, see related discussion https://github.com/minkphp/Mink/pull/826)

serializing elements is natively supported by W3C WebDriver [1] and already fully supported by `instaclick/php-webdriver` [2] since v1.4.13

[1] https://w3c.github.io/webdriver/#executing-script
[2] https://github.com/instaclick/php-webdriver/pull/110

example usage:

```
// expecting:
// a) $minkSession with working \WebDriver\Session driver
// b) loaded webpage with at least one <div> element with some ID attribute

$dumpElementFx = function ($v) {
    return $v instanceof \Behat\Mink\Element\NodeElement
        ? 'NodeElement: ' . $v->getXpath()
        : $v;
};

/** @var \Behat\Mink\Driver\Selenium2Driver */
$driver = $minkSession->getDriver();

$element = $driver->find('//div')[0]; // returns instance of \Behat\Mink\Element\NodeElement
var_dump($dumpElementFx($element));

$jsResult = $driver->evaluateScript(
    'return (function (elem) { return [elem, elem.id]; })(arguments[0])',
    [$element],
);

// will dump array with:
// [0] (new) instance of \Behat\Mink\Element\NodeElement to demonstrate conversion from php to webdriver and back
// [1] ID of given element extracted by JS
var_dump(array_map(fn ($v) => $dumpElementFx($v), $jsResult));

$jsResult = $driver->evaluateScript(
    'return (function (elem) { return [elem, elem.id]; })(arguments[0])',
    [$jsResult[0]],
);

// dump must be the same, but this time, element passed to a Selenium2Driver::evaluateScript had no xpath
var_dump(array_map(fn ($v) => $dumpElementFx($v), $jsResult));
```


src/Selenium2Driver.php
```diff
@@ -246,6 +249,68 @@ protected static function charToOptions($char, $modifier = null)
...
+     * @param mixed $data
+     *
+     * @return mixed
+     */
+    private function unserializeExecuteResult($data)
+    {
+        if ($data instanceof Element) {
+            $minkSession = \Closure::bind(function () {
+                return $this->session;
+            }, $this, CoreDriver::class)();
```
stof, line 296: this is a no-go. The private state of classes from a different package are not covered by BC. And we do have a (stale) work in progress changing the way NodeElement gets instantiated, which is precisely why the current API of CoreDriver ensures that driver classes don't need to handle that instantiation themselves.

src/Selenium2Driver.php
```diff
@@ -246,6 +249,68 @@ protected static function charToOptions($char, $modifier = null)
...
+
+            return new NodeElement(
+                self::XPATH_WEBDRIVER_ELEMENT_PREFIX
```
stof, line 299: would this generated XPATH work when passing this element to a different method of the driver ? Or is it something that you totally made up, and that would only work for passing it to other calls to `executeScript` ?

src/Selenium2Driver.php
mvorisek: `Selenium2Driver` is child class of `CoreDriver` class, thus I find this code quite reasonable as this is exactly what the `CoreDriver` class does https://github.com/minkphp/Mink/blob/97123e0048c3c9db56e153152b11a2d9ba643dbb/src/Driver/CoreDriver.php#L102

is this reasoning enough, or you want me to submit a PR to `behat/mink` to change the visibility of `CoreDriver::session` prop to `protected`? or do you have in mind a better solution?

src/Selenium2Driver.php
mvorisek: good point - I need to add support for other WebDriver methods, the artificial xpath is supposed to work with any method that does accept `NodeElement` (or 1:1 xpath from it)

src/Selenium2Driver.php
stof: We don't want to make it protected. That would force it to be covered by BC, which would prevent us from removing it from behat/mink when we finally do that refactoring (currently, all our drivers are compatible with the in-progress stale refactoring with no changes in them).

mvorisek: So what solution do you propose? If public `CoreDriver::find` method is supposed to keep returning `NodeElement`, I belive the session object will be always needed.

What about moving https://github.com/minkphp/Mink/blob/97123e0048c3c9db56e153152b11a2d9ba643dbb/src/Driver/CoreDriver.php#L102 to a protected method `CoreDriver::createNodeElementFromXpath`? This way, the session object can be kept private.

mvorisek: Even if there is no support for native xpath generation from JS/Selenium, given the fact how dummily the xpaths are generated in https://github.com/minkphp/MinkSelenium2Driver/blob/086c6a84093702f89123d3ce42c27d091d9dc16d/src/Selenium2Driver.php#L521 , I belive with one more roundtrip to Selenium, we can obtain xpath via short JS code. This will also solve https://github.com/minkphp/MinkSelenium2Driver/pull/352#discussion_r851288614 cleanly and the returned NodeElement will be like any other.

src/Selenium2Driver.php
stof: to me, adding hacks in all method to support a fake XPath is not the solution. Mink does some XPath manipulation when using `NodeElement::find()` for instance.
The XPath passed in NodeElement must be a valid XPath, not a hacked string. Otherwise, you will break Mink.

src/Selenium2Driver.php
stof: why saying this is dummy ? This is a way to produce the xpath for all those elements without doing more roundtrips to selenium.

mvorisek: because if any element matching given xpath is prepended, previously created `NodeElement` will match a different DOM element later - is that wanted?

src/Selenium2Driver.php
mvorisek: I completely agree, fixed, now xpath is build and NodeElement is created/usable like any other.

src/Selenium2Driver.php
mvorisek: completely reworked, see file changes and https://github.com/minkphp/MinkSelenium2Driver/pull/352#discussion_r854002660 discussion

mvorisek:
@stof I need to add a test, but can you please review this PR briefly?

PR #355, opened 2022-04-20
Pure refactoring - use nowdoc instead of heredoc

mvorisek:


mvorisek:
not supported by PHP 7.2

src/Selenium2Driver.php
```diff
@@ -585,44 +585,44 @@ public function getValue($xpath)
...
+                for (var i = 0; i < node.options.length; i++) {
+                    if (node.options[i].selected) {
+                        value.push(node.options[i].value);
+                    }
+                }
 
-return value;
-JS;
+                return value;
+                JS;
```
stof, line 625: this is a no-go. The driver supports PHP 7.2.

stof:
the issue is not about nowdoc vs heredoc, but about the indented variant of them.

mvorisek:
> the issue is not about nowdoc vs heredoc, but about the indented variant of them.

I find nowdoc safer as no variables interpolation is possibles, but the changed LOC is low when not reindent, thus I will submit it as a part of another PR

PR #356, opened 2022-04-28
fix: Check if cookies are iterable before iterating

das-peter:
It can happen that the returned value is not iterable. The geetAllCookies() methode also declares the its return as "mixed".
Just about having slightly more defensive code.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -462,17 +462,19 @@ class Selenium2Driver extends CoreDriver
     public function getCookie($name)
     {
         $cookies = $this->wdSession->getAllCookies();
-        foreach ($cookies as $cookie) {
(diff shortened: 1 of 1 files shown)


stof:
In which case does this happen ?

das-peter:
I had a request to `wd/hub/session/chrome-100-0-4896-75-1-bc5dc535-0653-4f64-814c-d172db4c1582/session/bed6779a89bb52104cef11e83d40d7e6/cookie` return `null` as value. 
Underlying library is `instaclick/php-webdriver` - `\WebDriver\Session::getAllCookies()` which just passes the value as received from the endpoint back to the consumer.
It is highly likely this is / was an error on the endpoint but I don't think this warrants a generic failure.

aik099 (approved):

PR #358, opened 2022-10-16
Upgrade JS syn to v0.15.0

### LINKED ISSUE #279: Queries re use of Syn library

mvorisek:
syn.js downloaded by `npm install syn` and moved from `node_modules/syn/dist/global/syn.js` (no local compilation, only non-minified version is available, thankfully, as it is much better for debugging)

replaces https://github.com/minkphp/MinkSelenium2Driver/pull/333

fixes #279 and maybe also #20

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -204,7 +204,7 @@ class Selenium2Driver extends CoreDriver
     protected function withSyn()
     {
         $hasSyn = $this->wdSession->execute(array(
-            'script' => 'return typeof window["Syn"]!=="undefined" && typeof window["Syn"].trigger!=="undefined"',
+            'script' => 'return window.syn !== undefined && window.syn.trigger !== undefined',
             'args'   => array()
         ));
 
@@ -222,25 +222,29 @@ class Selenium2Driver extends CoreDriver
     /**
      * Creates some options for key events
      *
-     * @param string $char     the character or code
-     * @param string $modifier one of 'shift', 'alt', 'ctrl' or 'meta'
+     * @param string|int  $char     the character or code
+     * @param string|null $modifier one of 'shift', 'alt', 'ctrl' or 'meta'
      *
      * @return string a json encoded options array for Syn
      */
     protected static function charToOptions($char, $modifier = null)
     {
-        $ord = ord($char);
-        if (is_numeric($char)) {
-            $ord = $char;
+        if (is_int($char)) {
+            $charCode = $char;
+            $char = chr($charCode);
+        } else {
+            $charCode = ord($char);
         }
 
         $options = array(
-            'keyCode'  => $ord,
-            'charCode' => $ord
+            'key'  => $char,
+            'which'  => $charCode,
+            'charCode'  => $charCode,
+            'keyCode'  => $charCode,
         );
 
         if ($modifier) {
-            $options[$modifier.'Key'] = 1;
+            $options[$modifier . 'Key'] = true;
         }
 
         return json_encode($options);
@@ -1167,7 +1171,7 @@ JS;
      */
     private function trigger($xpath, $event, $options = '{}')
     {
-        $script = 'Syn.trigger("' . $event . '", ' . $options . ', {{ELEMENT}})';
+        $script = 'syn.trigger({{ELEMENT}}, "' . $event . '", ' . $options . ')';
         $this->withSyn()->executeJsOnXpath($xpath, $script);
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+1.7.0 / 2023-02-26
+==================
+
+BC break:
+
+* `syn` JS library was upgraded from v0.0.3 to v0.15.0
+
+Bug fixes:
+
+* `Selenium2Driver::{keyPress, keyDown, keyUp}` methods were using numeric string input as char code. Newly an integer type must be passed.
+
 1.6.0 / 2022-03-28
 ==================
 
--- a/src/Resources/syn.js
+++ b/src/Resources/syn.js
@@ -1,8 +1,2954 @@
 /**
- * Syn - 0.0.3
+ * Syn - 0.15.0
  *
- * @copyright 2014 Bitovi
- * Mon, 30 Jun 2014 22:44:59 GMT
+ * how to update:
+ * 1. npm install syn
+ * 2. copy the file from node_modules/syn/dist/global/syn.js
+ * 3. adjust the header to match the installed version
+ *
+ * @copyright Bitovi
  * @license MIT
  */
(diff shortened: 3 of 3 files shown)


src/Selenium2Driver.php
```diff
@@ -222,25 +222,29 @@ protected function withSyn()
...
      *
-     * @param string $char     the character or code
-     * @param string $modifier one of 'shift', 'alt', 'ctrl' or 'meta'
+     * @param string|int  $char     the character or code
+     * @param string|null $modifier one of 'shift', 'alt', 'ctrl' or 'meta'
      *
      * @return string a json encoded options array for Syn
      */
-    protected static function charToOptions($char, $modifier = null)
+    protected static function charToOptions($char, $modifier)
```
stof, line 230: turning a argument into a mandatory one in a public or protected method is a BC break.

src/Selenium2Driver.php
```diff
@@ -222,25 +222,29 @@ protected function withSyn()
     /**
      * Creates some options for key events
      *
-     * @param string $char     the character or code
-     * @param string $modifier one of 'shift', 'alt', 'ctrl' or 'meta'
+     * @param string|int  $char     the character or code
```
stof, line 225: why adding support for integers there ? This looks like a new feature

stof: This method is protected and not tagged as `@internal`. So someone extending the class (and I know some people do extend this class) could rely on it.

src/Selenium2Driver.php
mvorisek: I belive the default value here is not needed, but I reverted this change, it is not needed

src/Selenium2Driver.php
mvorisek: https://github.com/minkphp/MinkSelenium2Driver/pull/358#discussion_r997034281 resolved, the phpdoc is changed correctly and `int` can/is passed by the testing suite

stof: if integers are passed without being handled explicitly here, how did it work before ? Were we dispatching broken events (in which case it mean that this is not _actually_ tested by the driver testsuite as it was passing) or was the old version of Syn fixing this for us ?

src/Selenium2Driver.php
```diff
@@ -204,7 +204,7 @@ public static function getDefaultCapabilities()
     protected function withSyn()
     {
         $hasSyn = $this->wdSession->execute(array(
-            'script' => 'return typeof window["Syn"]!=="undefined" && typeof window["Syn"].trigger!=="undefined"',
+            'script' => 'return typeof window.syn !== \'undefined\' && typeof window.syn.trigger !== \'undefined\'',
```
aik099, line 207: Replacing `typeof ... !== \'undefined\'` constructs with `... !== undefined` would also work.

src/Selenium2Driver.php
aik099: @stof , I can confirm, that:

* `\Behat\Mink\Driver\DriverInterface::keyPress` method PhpDoc has `@param string|int $char     ...`;
* implementation of that interface method `\Behat\Mink\Driver\Selenium2Driver::keyPress` is passing that `$char` parameter to the `\Behat\Mink\Driver\Selenium2Driver::charToOptions` method (changed below).

I agree with the made PhpDoc change, but I don't agree with the `is_numeric` change into `is_int` as this could be a BC break.

src/Selenium2Driver.php
```diff
@@ -222,25 +222,29 @@ protected function withSyn()
...
     {
-        $ord = ord($char);
-        if (is_numeric($char)) {
-            $ord = $char;
+        if (is_int($char)) {
+            $charCode = $char;
+            $char = chr($charCode);
+        } else {
+            $charCode = ord($char);
         }
```
aik099, line 237: Please revert this code fragment, because it could potentially lead to the BC break when `is_int($char)` and `is_numeric($char)` yields different results.

aik099 (changes requested):

src/Selenium2Driver.php
mvorisek: @aik099 for ex. when `1` char/string is passed, I expect `1` to be pressed, `is_numeric` was wrong here, also all tests pass

aik099: That's an interesting observation. Never thought that number (e.g. `1`, `7`) could also be interpreted as a symbol and not a char code. Need to look closely at these tests you're talking about.

mvorisek: @aik099 please merge https://github.com/minkphp/driver-testsuite/pull/61 test for it, also adds a test for capital letter key press

mvorisek: @aik099 can you please merge the test PR and this PR then?

aik099: @stof , I've debugged the original (prior to this PR) version of the `\Behat\Mink\Driver\Selenium2Driver::charToOptions` method and the results are as follows:

* When a letter is given as input, then it's converted into an ASCII code
* When a number is given (as a string or as an integer) it's interpreted as an ASCII code itself

This behavior partially matches the DocBlock on the `\Behat\Mink\Driver\DriverInterface::keyPress`:

```
* @param string|int $char     could be either char ('b') or char-code (98)
 ```

Due to the usage of the `is_numeric` function (in implementation), a digit from `'0'` to `'9'` passed in as a string was interpreted as an ASCII code as well, which is a bug.

To overcome this developers likely had to pass in the ASCII code of these digits. Tests for the such case were missing, but now they're added via https://github.com/minkphp/driver-testsuite/pull/61 .

File: src/Selenium2Driver.php (hunk 2 above)
Comment (aik099) [reply src/Selenium2Driver.php:225]: After a thorough investigation, I longer consider this a BC break and agree with a merge.

**Comment by aik099** [comment]
@mvorisek , I've rerun all tests with an updated test suite and now all `key*` method tests fail due some kind of missing space:

```
-'key pressed: 82 / 0 / 0 / 0 / 1'
+'key pressed:82 / 0 / 0 / 0 / 1'
```

Please investigate. Not sure of this is connected to changes in the driver test suite or this PR.

mvorisek:
@aik099 yes, please merge this PR as well and close the 3 mentioned #/issues in the description then.

aik099:
> @aik099 yes, please merge this PR as well and close the 3 mentioned #/issues in the description then.

I'll wait for @stof to see if he has anything to all.

mvorisek:
Hi @stof, this PR is approved but Alex wants your feedback as well, can you please give this PR a green light?

mvorisek:
@stof @aik099 can this PR be merged? Thank you.

aik099:
@mvorisek , I guess you need to add a line to the changelog explaining the bug, that was fixed with this PR: the numbers from `0` to `9` passed as strings were treated as ASCII codes.

After that I'll try to merge this PR.

mvorisek:
@aik099 CHANGELOG added

aik099 (approved):

aik099:
Merged. Thank you @mvorisek .

mvorisek:
@aik099 Thank you, would it be possible to tag a new release please?

aik099:
How that would help you?

mvorisek:
I no longer need it as I desided to fork this repo..

But I would like to get my changes still merged, especially https://github.com/minkphp/MinkSelenium2Driver/pull/359 and https://github.com/minkphp/driver-testsuite/pull/59 . Thank you.

PR #354, opened 2022-04-18
Run tests also with Selenium 3.x

mvorisek:
There were 3 errored and 2 failing tests, see post below for analysis and how they were fixed.

PR #359, opened 2022-10-17
No WebDriver\Session::moveto() call when dragging onto itself

mvorisek:
This fixes dragging over itself, ie. source element = destination element, in such usecase, extra `WebDriver\Session::moveto()` call is redundant.

Drag library like https://github.com/Shopify/draggable needs this fix as the source element is very often transformed when dragging and `WebDriver\Session::moveto()` is failing otherwise.

image

also remove `Selenium2Driver::withSyn()` call as completely useless for dragging, the current impl. of `Selenium2Driver::dragTo()` does not use it

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -981,9 +981,11 @@ JS;
         $this->executeJsOnElement($source, $script);
 
         $this->getWebDriverSession()->buttondown();
-        $this->getWebDriverSession()->moveto(array(
-            'element' => $destination->getID()
-        ));
+        if ($destination->getID() !== $source->getID()) {
+            $this->getWebDriverSession()->moveto(array(
+                'element' => $destination->getID()
+            ));
+        }
         $this->getWebDriverSession()->buttonup();
 
         $script = <<<JS
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -978,7 +978,7 @@ JS;
     element.dispatchEvent(event);
 }({{ELEMENT}}));
 JS;
-        $this->withSyn()->executeJsOnElement($source, $script);
+        $this->executeJsOnElement($source, $script);
 
         $this->getWebDriverSession()->buttondown();
         $this->getWebDriverSession()->moveto(array(
@@ -996,7 +996,7 @@ JS;
     element.dispatchEvent(event);
 }({{ELEMENT}}));
 JS;
-        $this->withSyn()->executeJsOnElement($destination, $script);
+        $this->executeJsOnElement($destination, $script);
     }
 
     public function executeScript(string $script)


stof:
Well, if we don't have a test covering it, we risk regressing on that case.
And if this is necessary to be compatible with the behavior of the draggable library of Shopify, it means there is a real-world use case for that, not just an implementation detail.

mvorisek:
if this will imply the fix will be merged then, I can code the test

> probably by hiding the dragged element on `dragstart` event

simply like this or use native https://github.com/Shopify/draggable lib? (about 100KB)

stof:
if we can do it with a simple implementation, it will be better for the driver testsuite IMO.

and yes, once we have tests for it, I would happily accept that patch.

mvorisek:
@stof please merge, tested in https://github.com/minkphp/driver-testsuite/pull/59

I will then address https://github.com/minkphp/driver-testsuite/pull/59#issuecomment-1288109277 and https://github.com/minkphp/MinkSelenium2Driver/pull/354

mvorisek:
Hi @stof, can we finish this PR?

mvorisek:
How to skip the `Behat\Mink\Tests\Driver\Js\JavascriptTest::testDragDropOntoHiddenItself` test on Selenium 2? Or any idea how to fix the driver for Selenium 2?

aik099:
@mvorisek , Test skipping can be done in the `\Behat\Mink\Tests\Driver\Selenium2Config::skipMessage` method.

The only purpose of the test you've mentioned (added by you) was to confirm, that code changes from this PR (added by you) will continue to work as expected.

Then, why do you need that test to be skipped?

mvorisek:
In Selenium 3 (and with regular mouser/keyboard) the dragging works. In Selenium 2, it seems it does not, IDK why. Help welcomed. I did some experiments with Selenium 2 and it seems the dragging is not emulated well enough.

Please advise what to do and how to detect Selenium 2. Or how to fix Selenium 2 of course if you have an idea.

aik099:
I've checked on Selenium 2 + Chrome and it works, but Selenium 2 + Firefox doesn't. Visually it looks like once the element is clicked without releasing a mouse button (to perform drag) it disappears somewhere and never drops back.

Maybe it is a bug in Selenium or geckodriver itself.

aik099:
I've checked with different Firefox versions on Selenium 2 and :

* on Firefox 47 it doesn't work (used in GitHub Actions)
* on Firefox 92 it does work

I wish we could use more recent browsers on GitHub Actions, but I have no idea how to make this happen.

aik099:
@uuf6429 , do you know what tag of the `selenium-firefox` Dockerimage needs to be used to get the Selenium 2 and modern Firefox version?

Tests for this PR are failing only because of outdated Firefox version usage in the Selenium Dockerimage.

The same test failure also occurs on the https://github.com/minkphp/webdriver-classic-driver/pull/27 .

uuf6429:
@aik099 `selenium/standalone-firefox:2` typically means "the latest selenium 2 with the latest firefox available at that time".
In other words, I don't think they are building images for with selenium 2 and updated firefox since they stopped maintaining selenium 2.

That being said, we could probably reverse engineer it (the docker layers are publicly readable) to have an up-to-date firefox.

aik099:
> @aik099 `selenium/standalone-firefox:2` typically means "the latest selenium 2 with the latest firefox available at that time". In other words, I don't think they are building images for with selenium 2 and updated firefox since they stopped maintaining selenium 2.
> 
> That being said, we could probably reverse engineer it (the docker layers are publicly readable) to have an up-to-date firefox.

If that can be done, then both Selenium-based drivers would benefit from this. I've created a #391 about this.

aik099:
Merging, because test from the https://github.com/minkphp/driver-testsuite/pull/59 is already merged. Thanks @mvorisek .

PR #361, opened 2023-02-14
Iframe switch - Send element info as selenium 4 require it

omarlopesino:
## Problem

When I try to switch to an iframe with the switchToIframe method I receive the following error:

```bash
invalid argument: 'id' can not be string
```
## Steps to reproduce

Call the method 'switchToIframe' method of the Selenium2DriverClass passing the 'id' of the iframe it is needed to switch to as a string:

```
  $driver->switchToIframe('iframe-id');
```

## Proposed resolution
According to the [current specs](https://www.w3.org/TR/webdriver1/#switch-to-frame) , it must be an object that contains the element ID inside. So, in the switchToIframe method, we must change the format to send the element ID (UUID) of the element we are switching to, instead of the plain id passed by arguments. When the name of the iframe passed is null, we must send nothing.

Currently, this solution makes the library not compatible with selenium 3, but as far as I understand selenium 4 is stable so it should be replaced.

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -430,7 +430,7 @@ class Selenium2Driver extends CoreDriver
           ];
         }
         else {
-          $element_info = NULL;
+          $element_info = null;
         }
 
         $this->wdSession->frame(array('id' => $element_info));
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -424,7 +424,7 @@ class Selenium2Driver extends CoreDriver
     public function switchToIFrame($name = null)
     {
         if (!empty($name)) {
-          $element = $this->wdSession->element('id', $name);
+          $element = $this->wdSession->element('name', $name);
           $element_info = [
             'ELEMENT' => $element->getID(),
           ];
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -423,7 +423,17 @@ class Selenium2Driver extends CoreDriver
      */
     public function switchToIFrame($name = null)
     {
-        $this->wdSession->frame(array('id' => $name));
+        if (!empty($name)) {
+          $element = $this->wdSession->element('id', $name);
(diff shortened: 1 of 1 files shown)


aik099:
> Currently, this solution makes the library not compatible with selenium 3, but as far as I understand selenium 4 is stable so it should be replaced.

Still a BC break.

src/Selenium2Driver.php
```diff
@@ -423,7 +423,17 @@ public function switchToWindow($name = null)
      */
     public function switchToIFrame($name = null)
     {
-        $this->wdSession->frame(array('id' => $name));
+        if (!empty($name)) {
+          $element = $this->wdSession->element('name', $name);
+          $element_info = [
+            'ELEMENT' => $element->getID(),
```
mvorisek, line 429: Thank you for this PR, it will solve https://github.com/minkphp/driver-testsuite/pull/60

`'ELEMENT'` key is not compatible with W3C spec, see https://github.com/instaclick/php-webdriver/commit/b5b18bd2e02b35a14c7085552780588b8876460b

Will tests pass if you pass `$element` directly like `$this->wdSession->frame(array('id' => $element))`?

aik099:
The frame switching support (works for both Selenium 2 and 3) was implemented in https://github.com/minkphp/MinkSelenium2Driver/pull/382.

For Selenium 4 support use the https://github.com/minkphp/webdriver-classic-driver driver.

Thank you, @mvorisek .

src/Selenium2Driver.php
aik099: > Will tests pass if you pass `$element` directly like `$this->wdSession->frame(array('id' => $element))?`

No, tests will fail, because `instaclick/php-webdriver` doesn't support this. The https://github.com/instaclick/php-webdriver/pull/132 should add this functionality.

In the meanwhile the workaround was created in the https://github.com/minkphp/MinkSelenium2Driver/pull/382.

Issue #193, opened 2014-12-22
document's title has different content ?


kendoctor:
//Given this element is a title element
$element->getText(); // with blank content
$element->getHtml(); // it's ok 
1. I put chinese characters in title  : <title>活动首页</title>
2. I used selenium driver to test it.

Selenium Driver :

``` php
    public function getText($xpath)
    {
        $node = $this->findElement($xpath);
        $text = $node->text();
        $text = (string) str_replace(array("\r", "\r\n", "\n"), ' ', $text);

        return $text;
    }

    public function getHtml($xpath)
    {
        return $this->executeJsOnXpath($xpath, 'return {{ELEMENT}}.innerHTML;');
    }
```

Why these two methods have different implementations for finding element?
Why not to convert getHtml() returned result to text ?

PR #362, opened 2023-02-17
Use "innerText" attribute in getText() method

yyss8:
Hi, we recently moved the tests of our Drupal site from using [chrome-mink-driver](https://gitlab.com/behat-chrome/chrome-mink-driver) to `MinkSelenium2Driver` and noticing some of the tests are failing due to an empty string being returned when trying to assert the title of the page.

We found a similar issue here https://github.com/minkphp/MinkSelenium2Driver/issues/193 and it seems to be an upstream issue with Selenium? 

It also lead us to this [stackoverflow answer](https://stackoverflow.com/questions/20888592/gettext-method-of-selenium-chrome-driver-sometimes-returns-an-empty-string#answer-20963084) and which recommends to use `innerText` attribute to retrieve the text of the element.

Similar issue reported on [Drupal.org](https://www.drupal.org/project/drupal/issues/3025845)

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -538,8 +538,8 @@ class Selenium2Driver extends CoreDriver
     public function getText($xpath)
     {
         $node = $this->findElement($xpath);
-        $text = $node->text();
-        $text = (string) str_replace(array("\r", "\r\n", "\n"), ' ', $text);
+        $text = $node->attribute('innerText');
+        $text = trim(preg_replace('/\s+/u', ' ', $text), ' ');
 
         return $text;
     }
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -539,7 +539,7 @@ class Selenium2Driver extends CoreDriver
     {
         $node = $this->findElement($xpath);
         $text = $node->attribute('innerText');
-        $text = trim(preg_replace('/\s+/u', ' ', $text), ' ');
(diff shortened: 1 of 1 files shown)


aik099:
@yyss8 can you please reference the issue on the Selenium side?

I don't understand what Selenium the bug prevents the `->text()` method from working on some nodes.

mvorisek:
I belive the problem is https://github.com/minkphp/MinkSelenium2Driver/pull/327. I would be happy @aik099 if my original PR can be reconsidered, as the `getText()` will be than implemented consistently like `getHtml()`.

aik099:
@yyss8 , would it be possible to rewrite your tests to perform actions/assertions based on user possible actions in the Web Browser (e.g. not checking the text of the elements, that aren't visible on the page)?

mvorisek:
Here is how I fixed it in my fork: https://github.com/atk4/behat-mink-selenium2-driver/commit/05e2230c35dc573004880bfd2107d660c7b4790a

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText `innerText` property is well defined:
> As a getter, it approximates the text the user would get if they highlighted the contents of the element with the cursor and then copied it to the clipboard.

> If the element itself is not [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered) (for example, is detached from the document or is hidden from view), the returned value is the same as the [Node.textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) property.

PR #364, opened 2023-05-01
Fix syn.js v0.15.0 input event to be synchronous

mvorisek:
based on https://github.com/bitovi/syn/pull/196, fixes https://github.com/bitovi/syn/issues/120

input event must be synchronous per spec - https://w3c.github.io/uievents/#event-type-input

repro case: https://github.com/bitovi/syn/issues/120#issuecomment-1529794347


aik099:
@mvorisek , do I understand correctly, that you're proposing the Syn library change to Mink, which isn't yet merged into the Syn library itself?

PR #365, opened 2023-05-10
Create LICENSE.md

jeremychinquistdrunomicscom:
Our project requires that the MIT license text be added to the repository. I copied it from the https://spdx.org/licenses/MIT.html#licenseText


aik099:
Can you copy/paste https://github.com/minkphp/driver-testsuite/blob/master/LICENSE file into this repo and update names/years in it? Then it would be more consistent.

jeremychinquistdrunomicscom:
@aik099 
I did that but made one exception. I left "MIT License" at the top because it is an indication again of the license and this is used elsewhere. 
Should Konstantin also be added, or simply Pete?
Please merge if this is enough.

aik099:
> I did that but made one exception. I left "MIT License" at the top because it is an indication again of the license and this is used elsewhere.

I've removed it just because other driver repos don't have this heading and GitHub is detecting license type already (either by it's text or `MIT` word in the `composer.json`.

> Should Konstantin also be added, or simply Pete?

I'm not sure. I've left it as is.

aik099:
@jeremychinquistdrunomicscom , merging. Thank you.

jeremychinquistdrunomicscom:
Wonderful, thank you.

PR #372, opened 2023-10-03
W3C Compatability Fixes

### LINKED ISSUE #293: Support the W3C webdriver spec

justafish:
Fixes to enable compatibility with WebDriver.

Here are the modifications running against the full set of Drupal core tests: https://github.com/Lullabot/drupal-webdriver/pull/2

To run against newer browsers/versions of selenium that have dropped support for json wire, a drop-in replacement library for instaclick/php-webdriver is needed (this is intended to null any disruption for existing test setups, but projects using instaclick/php-webdriver are currently stuck on selenium versions over a year old). Here's the replacement library, along with these proposed changes, run against the Drupal core tests as above: https://github.com/Lullabot/drupal-webdriver/pull/1

Related drupal.org issue https://www.drupal.org/project/drupal/issues/3240792
Fixes #293

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -78,11 +78,9 @@ jobs:
           mkdir ./logs
           ./vendor/bin/mink-test-server &> ./logs/mink-test-server.log &
 
-      # To run this locally, additionally mount a volume with the repo to retrieve test files
-      # e.g. -v /home/youruser/MinkSelenium2Driver:/home/youruser/MinkSelenium2Driver
       - name: Start Selenium
         run: |
-          docker run --net host --name selenium --volume /dev/shm:/dev/shm --volume $GITHUB_WORKSPACE:$GITHUB_WORKSPACE --shm-size 2g selenium/standalone-firefox:${{ matrix.selenium }} &> ./logs/selenium.log &
+          docker run --net host --name selenium --volume /dev/shm:/dev/shm --volume $GITHUB_WORKSPACE:$GITHUB_WORKSPACE -e SE_NODE_OVERRIDE_MAX_SESSIONS=true -e SE_NODE_MAX_SESSIONS=5 --shm-size 2g selenium/standalone-firefox:${{ matrix.selenium }} &> ./logs/selenium.log &
 
       - name: Wait for browser & PHP to start
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -66,6 +66,13 @@ jobs:
         run: |
           composer update --no-interaction --prefer-dist
 
+      - name: Replace instaclick/php-webdriver with lullabot/php-webdriver for new selenium versions
+        run: |
+          if [ "${{ matrix.selenium }}" == "latest" ]; then
+            # @todo replace this with a release
+            composer require lullabot/php-webdriver:dev-main
+          fi
+
       - name: Setup Mink test server
         run: |
           mkdir ./logs
@@ -80,19 +87,12 @@ jobs:
           while ! nc -z localhost 4444 </dev/null; do echo Waiting for remote driver to start...; sleep 1; done
           while ! nc -z localhost 8002 </dev/null; do echo Waiting for PHP server to start...; sleep 1; done
 
-      - name: Replace instaclick/php-webdriver with lullabot/php-webdriver for new selenium versions
-        run: |
-          if [ "${{ matrix.selenium }}" == "latest" ]; then
-            # @todo replace this with a release
-            composer require lullabot/php-webdriver:dev-main
-          fi
-
       - name: Run tests
         run: |
           if [ "${{ matrix.selenium }}" == "latest" ]; then
             DRIVER_URL="http://localhost:4444" ./vendor/bin/phpunit -v --coverage-clover=coverage.xml
           else
-            DRIVER_URL="http://localhost:4444/wd/hub" ./vendor/bin/phpunit -v --coverage-clover=coverage.xml
+            ./vendor/bin/phpunit -v --coverage-clover=coverage.xml
           fi
 
       - name: Upload coverage
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -73,16 +73,17 @@ jobs:
 
       - name: Start Selenium
         run: |
-          docker run --net host --name selenium --volume /dev/shm:/dev/shm --shm-size 2g selenium/standalone-firefox:"${{ matrix.selenium }}" &> ./logs/selenium.log &
+          docker run --net host --name selenium --volume /dev/shm:/dev/shm --shm-size 2g selenium/standalone-firefox:${{ matrix.selenium }} &> ./logs/selenium.log &
 
       - name: Wait for browser & PHP to start
         run: |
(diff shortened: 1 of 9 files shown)


justafish:
tests were initially failing as `isW3C()` is only available in the main branch of instaclick/php-webdriver https://github.com/instaclick/php-webdriver/blob/master/lib/WebDriver/Session.php#L136

stof:
See https://github.com/minkphp/webdriver-classic-driver for our in-progress work on implementing a new WebDriver-based driver for Mink (for now, all the initial implementation is in a PR, not yet in the main branch).

justafish:
@stof ah interesting :smile_cat: :+1: 

Would you please consider accepting this change into this project in the meantime? It's a small update, not disruptive to existing users, is successfully running over Drupal core's extensive tests, and will allow projects to get off > 1 year old browsers for testing without waiting for the new project to be finished and released

src/Selenium2Driver.php
```diff
@@ -214,6 +214,23 @@ public static function getDefaultCapabilities()
...
+    /**
+     * Checks if the WebDriver session is W3C compatible.
+     *
+     * @return bool
+     *
+     * @throws UnknownCommand
+     */
+    public function isW3C() {
+        try {
+            // This method doesn't exist in older versions of php-webdriver.
```
stof, line 226: I'd rather check with `method_exists` here rather than relying on the undefined method call triggering an UnknownCommand exception.

src/Selenium2Driver.php
```diff
@@ -214,6 +214,23 @@ public static function getDefaultCapabilities()
         );
     }
 
+    /**
+     * Checks if the WebDriver session is W3C compatible.
+     *
+     * @return bool
+     *
+     * @throws UnknownCommand
```
stof, line 222: This is wrong. This method will never throw this exception (it precisely catches it)

src/Selenium2Driver.php
```diff
@@ -921,11 +971,20 @@ public function wait(int $timeout, string $condition)
...
     {
-        $window = $this->getWebDriverSession()->window($name ?: 'current');
-        \assert($window instanceof Window);
-        $window->postSize(
-            array('width' => $width, 'height' => $height)
-        );
+        if ($this->isW3C()) {
+            // This method doesn't exist in older versions of php-webdriver.
+            /** @phpstan-ignore-next-line variable.undefined */
+            $this->wdSession->window($name ? $name : 'current')->postRect(
```
stof, line 977: this must use `getWebDriverSession()` to have the proper error handling if the method is called on a non-started driver instead of crashing by calling a method on `null`

.github/workflows/tests.yml
```diff
@@ -71,16 +73,26 @@ jobs:
...
 
       - name: Wait for browser & PHP to start
         run: |
           while ! nc -z localhost 4444 </dev/null; do echo Waiting for remote driver to start...; sleep 1; done
           while ! nc -z localhost 8002 </dev/null; do echo Waiting for PHP server to start...; sleep 1; done
 
+      - name: Replace instaclick/php-webdriver for new selenium versions
+        run: |
+          if [ "${{ matrix.selenium }}" == "latest" ]; then
+            composer require lullabot/php-webdriver:dev-main
```
stof, line 86: This published fork does not publish the fact that it replaces `instaclick/php-webdriver` so composer will happily install both in parallel, with no guarantee about which one will win for defining the class. So this looks totally wrong to me.
In the current situation, I'm against merging this PR and releasing as is, because it actually makes the package unreliable for users.

justafish: @stof it is marked as replacing it https://github.com/Lullabot/php-webdriver/blob/main/composer.json#L19 and I can see it doing so in the tests I just added:
https://github.com/minkphp/MinkSelenium2Driver/actions/runs/6392945083/job/17351217167?pr=372
```
Run if [ "latest" == "latest" ]; then
  if [ "latest" == "latest" ]; then
    composer require lullabot/php-webdriver:dev-main
  fi
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    COMPOSER_PROCESS_TIMEOUT: 0
    COMPOSER_NO_INTERACTION: 1
    COMPOSER_NO_AUDIT: 1
./composer.json has been updated
Running composer update lullabot/php-webdriver
Loading composer repositories with package information
Updating dependencies
Lock file operations: 1 install, 0 updates, 1 removal
  - Removing instaclick/php-webdriver (1.4.16)
  - Locking lullabot/php-webdriver (dev-main 3014d58)
Writing lock file
Installing dependencies from lock file (including require-dev)
Package operations: 1 install, 0 updates, 1 removal
  - Downloading lullabot/php-webdriver (dev-main 3014d58)
  - Removing instaclick/php-webdriver (1.4.16)
  - Installing lullabot/php-webdriver (dev-main 3014d58): Extracting archive
Package phpunit/php-token-stream is abandoned, you should avoid using it. No replacement was suggested.
Generating autoload files
26 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
```

I'll make a proper release of the library before this new test is merged, I just wanted to get tests etc passing here first :smile:

stof: ah indeed. By default, packagist shows the info for the latest release, and this is only in the dev version.

justafish:
@stof ready for review!
I'll wait on making a release of lullabot/php-webdriver until this PR is happy

alexpott:
@justafish this looks great. It'd be nice to run Drupal's tests against firefox and other browsers at last. @stof is there scope to merge this as a stop gap until the work on https://github.com/minkphp/webdriver-classic-driver is complete?

justafish:
Hi @stof - could you please let me know if the project is willing to accept this PR so I can make a release of php-webdriver? We can move Drupal to a fork of this package if necessary, but I would prefer to continue collaborating here :smile_cat:

andypost:
Is this PR still make sense?

aik099:
If https://github.com/minkphp/webdriver-classic-driver solves this PR needs, then probably not.

PR #383, opened 2024-02-23
Also test on Selenium 3

aik099:
## Observations

Used `selenium/standalone-firefox` Docker image chooses the Firefox version like so:

* Selenium 2.53.1 = Firefox 47
* Selenium 3.141.59 = Firefox 92

Not sure why, but some of the below tests fail only on Firefox 92.

## Used build

https://github.com/minkphp/MinkSelenium2Driver/actions/runs/8015977437/job/21897049068?pr=383

## Errors

> 1) Behat\Mink\Tests\Driver\Basic\IFrameTest::testIFrame
WebDriver\Exception\InvalidArgument: data did not match any variant of untagged enum FrameId at line 3 column 1
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:25:53'
System info: host: 'fv-az779-213', ip: '10.1.0.12', os.name: 'Linux', os.arch: 'amd64', os.version: '5.15.0-1056-azure', java.version: '1.8.0_[29](https://github.com/minkphp/MinkSelenium2Driver/actions/runs/8015977437/job/21897049622?pr=383#step:8:30)2'
Driver info: driver.version: unknown

* https://github.com/minkphp/MinkSelenium2Driver/pull/382

> 2) Behat\Mink\Tests\Driver\Js\ChangeEventTest::testSetValueChangeEvent with data set "file" ('the-file', '/home/runner/work/MinkSeleniu...e1.txt', '/home/runner/work/MinkSeleniu...e2.txt')
WebDriver\Exception\InvalidArgument: File not found: /home/runner/work/MinkSelenium2Driver/MinkSelenium2Driver/vendor/mink/driver-testsuite/web-fixtures/file1.txt
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:25:53'
System info: host: 'fv-az697-197', ip: '10.1.0.44', os.name: 'Linux', os.arch: 'amd64', os.version: '5.15.0-1056-azure', java.version: '1.8.0_292'
Driver info: driver.version: unknown

* https://github.com/minkphp/MinkSelenium2Driver/pull/385

> 3) Behat\Mink\Tests\Driver\Js\WindowTest::testWindow
WebDriver\Exception\NoSuchWindow: Unable to locate window: 
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:25:53'
[...]

PR #384, opened 2024-02-23
Fixing window switching on Selenium 3

aik099:
Fixes this error from the https://github.com/minkphp/MinkSelenium2Driver/pull/383:

```
Behat\Mink\Tests\Driver\Js\WindowTest::testWindow
WebDriver\Exception\NoSuchWindow: Unable to locate window:
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:25:53'
System info: host: 'fv-az697-197', ip: '10.1.0.44', os.name: 'Linux', os.arch: 'amd64', os.version: '5.15.0-10[56](https://github.com/minkphp/MinkSelenium2Driver/actions/runs/8015977437/job/21897049068?pr=383#step:8:57)-azure', java.version: '1.8.0_292'
Driver info: driver.version: unknown
```

Implementation ported from the https://github.com/minkphp/webdriver-classic-driver.

What's included:

1. [fixed] switching to window by name on Selenium 3
2. [fixed] window resizing and maximizing for a non-current window
3. [added] closing all, but the main window during the session reset
4. [changed] the `getWindowName` and `getWindowNames` methods to return window names instead of window handles (if the name is empty, then the handle is returned)

Related issues:

* https://github.com/minkphp/driver-testsuite/issues/94
* https://github.com/minkphp/driver-testsuite/pull/96

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -410,7 +410,20 @@ class Selenium2Driver extends CoreDriver
 
     public function reset()
     {
-        $this->getWebDriverSession()->deleteAllCookies();
+        $webDriverSession = $this->getWebDriverSession();
+
+        // Close all windows except the initial one.
+        foreach ($webDriverSession->window_handles() as $windowHandle) {
+            if ($windowHandle === $this->initialWindowHandle) {
+                continue;
+            }
+
+            $webDriverSession->focusWindow($windowHandle);
+            $webDriverSession->deleteWindow();
+        }
+
+        $this->switchToWindow();
+        $webDriverSession->deleteAllCookies();
     }
 
     public function visit(string $url)
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -31,6 +31,8 @@ use WebDriver\Window;
  */
 class Selenium2Driver extends CoreDriver
 {
+    private const W3C_WINDOW_HANDLE_PREFIX = 'w3cwh:';
+
     /**
      * Whether the browser has been started
      * @var bool
@@ -64,6 +66,11 @@ class Selenium2Driver extends CoreDriver
      * @var array{script?: int, implicit?: int, page?: int}
      */
     private $timeouts = array();
+
+    /**
+     * @var string|null
+     */
+    private $initialWindowHandle = null;
 
     /**
      * @var Escaper
@@ -343,6 +350,7 @@ class Selenium2Driver extends CoreDriver
         $this->started = true;
 
         $this->applyTimeouts();
+        $this->initialWindowHandle = $this->getWebDriverSession()->window_handle();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Selenium2Driver.php
```diff
@@ -425,7 +427,45 @@ public function back()
...
+        if (strpos($name, self::W3C_WINDOW_HANDLE_PREFIX) === 0) {
+            return substr($name, strlen(self::W3C_WINDOW_HANDLE_PREFIX));
+        }
+
+        // ..otherwise check if any existing window has the specified name
+
+        $origWindowHandle = $this->getWebDriverSession()->window_handle();
+
+        try {
+            foreach ($this->getWebDriverSession()->window_handles() as $handle) {
```
mvorisek, line 457: should be deduplicated with self::getWindowNames()

src/Selenium2Driver.php
```diff
@@ -492,12 +532,29 @@ public function getScreenshot()
 
     public function getWindowNames()
     {
-        return $this->getWebDriverSession()->window_handles();
+        $origWindow = $this->getWebDriverSession()->window_handle();
+
+        try {
+            $result = array();
+            foreach ($this->getWebDriverSession()->window_handles() as $tempWindow) {
+                $this->getWebDriverSession()->focusWindow($tempWindow);
```
mvorisek, line 540: isn't there really a better solution - this can be quote slow with a lot of windows

aik099: Agreed.

But Selenium doesn't have a way of executing JavaScript (for getting a window name) in a window before I switch to it. See https://stackoverflow.com/questions/25760357/get-title-of-window-without-switching-in-selenium-webdriver .

Any ideas?

src/Selenium2Driver.php
aik099: The `self::getWindowNames` method iterates all windows and returns an array with window names only.

@mvorisek , How it can help in this method, that needs to figure out what window handle matches to a given window name?

P.S.
This method does fewer iterations (=selenium API calls) in the case when the searched window isn't the last opened.

src/Selenium2Driver.php
mvorisek: Then any solution switching based on window name seems suboptimal not only because of this, but also because the window names can be non-unique.

User should open a window, store the handle, and then use the handle to switch the window.

src/Selenium2Driver.php
mvorisek: I prefer to completely remove this hack - https://github.com/minkphp/MinkSelenium2Driver/pull/384#discussion_r1501840465.

src/Selenium2Driver.php
aik099: Do you know any website, that has lots (number of windows, that can cause any slow down of test suite) of opened windows, that is in needed of Selenium-based testing?

uuf6429: @mvorisek

1. removing the hack will break backward compatibility
2. it will be (theoretically) slower if the desired window is always at the end
3. a micro-optimization would be to reverse the window handles array, if we know that typically the window we're interested is at the end
4. end-users can make use of `getWindowName()` to be safe (and fast) - as you said in the end

@aik099 we could, however, decide to deprecate usages `windowName` concept and replace it `windowHandle` concept, just to make this clear.

uuf6429 (approved):

src/Selenium2Driver.php
aik099: I've created https://github.com/minkphp/MinkSelenium2Driver/issues/387 to preserve this discussion upon PR close/merge.

aik099:
@stof , please review. Ideally please review my other PRs as well: https://github.com/minkphp/MinkSelenium2Driver/pulls/aik099 .

src/Selenium2Driver.php
aik099: Resolving.

All further discussion about using window handles instead of names should happen in the https://github.com/minkphp/Mink/issues/868 .

src/Selenium2Driver.php
aik099: Resolving.

All further discussion about using window handles instead of names should happen in the https://github.com/minkphp/Mink/issues/868 .

PR #385, opened 2024-02-25
Fixes file upload test on Selenium 3

aik099:


uuf6429 (approved):
Keep in mind this might conflict with #383. Looks good otherwise.

aik099:
> Keep in mind this might conflict with #383. Looks good otherwise.

Thank you for the warning. We will see if GitHub is clever enough to perform interactive conflict resolution upon merge.

PR #389, opened 2024-03-02
Fixing clicking outside of a viewport issue

aik099:


src/Selenium2Driver.php
```diff
@@ -773,6 +776,14 @@ private function clickOnElement(Element $element): void
         $element->click();
     }
 
+    private function scrollIntoView(Element $element): void
+    {
+        $this->executeJsOnElement(
+            $element,
+            "arguments[0].scrollIntoView({ behavior: 'instant', block: 'end', inline: 'nearest' });"
```
stof, line 783: shouldn't we use `nearest` as well in the `block` direction ?

aik099: I don't use the `scrollIntoView` method myself. I've only copied the whole line from the PHP WebDriver (see https://github.com/php-webdriver/php-webdriver/blob/11923b4ba312fc9d137266ae67438be65d4d500f/lib/Remote/RemoteWebElement.php#L249 ). I trust that they know what's right here.

aik099:
Not merging yet, because I think we need to add a missing test in driver-testsuite.

PR #401, opened 2024-11-11
Skip "testDragDropOntoHiddenItself" test on Selenium 2 + Firefox

### LINKED ISSUE #391: Use latest Firefox version with Selenium 2 on GitHub Actions
The https://hub.docker.com/r/selenium/standalone-firefox image provided by Selenium only supports Firefox 47 when used with Selenium server 2.x versions. This causes problems in a test suite because a working code (e.g. dragging a div into itself in #359) fails due:

* either outdated Firefox version
* either outdated geckodriver version

Need to create an image with a modern Firefox/geckodriver version and use it on GitHub Actions.

aik099:
Closes #391

### DIFF
--- a/tests/Selenium2Config.php
+++ b/tests/Selenium2Config.php
@@ -62,6 +62,18 @@ class Selenium2Config extends AbstractConfig
             return 'Checking status code is not supported.';
         }
 
+        if (
+            'Behat\Mink\Tests\Driver\Js\JavascriptTest' === $testCase
+            && 'testDragDropOntoHiddenItself' === $test
+        ) {
+            $seleniumVersion = $_SERVER['SELENIUM_VERSION'] ?? null;
+            $browser = $_SERVER['WEB_FIXTURES_BROWSER'] ?? null;
+
+            if ($seleniumVersion && version_compare($seleniumVersion, '3.0.0', '<') && $browser === 'firefox') {
+                return 'The Firefox browser compatible with Selenium Server 2.x doesn\'t fully implement drag-n-drop support.';
+            }
+        }
+
         return parent::skipMessage($testCase, $test);
     }
 
--- a/tests/Selenium2Config.php
+++ b/tests/Selenium2Config.php
@@ -7,6 +7,7 @@ use Behat\Mink\Driver\Selenium2Driver;
 use Behat\Mink\Tests\Driver\Basic\BasicAuthTest;
 use Behat\Mink\Tests\Driver\Basic\HeaderTest;
 use Behat\Mink\Tests\Driver\Basic\StatusCodeTest;
+use Behat\Mink\Tests\Driver\Js\JavascriptTest;
 
 class Selenium2Config extends AbstractConfig
 {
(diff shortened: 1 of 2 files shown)


tests/Selenium2Config.php
```diff
@@ -62,6 +62,18 @@ public function skipMessage($testCase, $test): ?string
             return 'Checking status code is not supported.';
         }
 
+        if (
+            'Behat\Mink\Tests\Driver\Js\JavascriptTest' === $testCase
+            && 'testDragDropOntoHiddenItself' === $test
```
uuf6429, line 67: Small hint: using something so allows IDEs and static code analysers to see this as a reference to the method (e.g. `[ctrl]`+`[click]`ing `testDragDropOntoHiddenItself` would open the method in PhpStorm):
```php
if ([JavascriptTest::class, 'testDragDropOntoHiddenItself'] === [$testCase, $test]) {
```

tests/Selenium2Config.php
```diff
@@ -62,6 +62,18 @@ public function skipMessage($testCase, $test): ?string
...
+        ) {
+            $seleniumVersion = $_SERVER['SELENIUM_VERSION'] ?? null;
+            $browser = $_SERVER['WEB_FIXTURES_BROWSER'] ?? null;
```
uuf6429, line 70: Shouldn't we use `getenv()` in those two cases?

uuf6429:
Other than the two minor comments I left, what about that failing build? I haven't investigated it, but at first look it doesn't sound random.

tests/Selenium2Config.php
aik099: Done.

tests/Selenium2Config.php
aik099: Please see https://github.com/minkphp/webdriver-classic-driver/issues/44 . 

To summarize:

1. when we use `<env...` in `phpunit.xml` then:
   * using `getenv` in PHP works
   * overriding from GitHub Actions works
   * using `$_SERVER` doesn't work
2. when we use `<server...` in `phpunit.xml` then:
   * using `getenv` in PHP doesn't work
   * overriding from GitHub Actions doesn't work
   * using `$_SERVER` works
3. when we don't use `<env...` or `<server...` in `phpunit.xml` then:
   * using `getenv` in PHP - ??? (haven't tested)
   * overriding from GitHub Actions works
   * using `$_SERVER` works

Ideally, we should switch to `<env` + `getenv` in the whole test suite, but that is BC break for drivers. Still waiting for @stof response in the above-mentioned issue.

uuf6429: Understood, too bad.

uuf6429 (approved):
Other than the failing test, this looks fine to me.

tests/Selenium2Config.php
aik099: I'll wait for a resolution on the above-mentioned issue. Maybe this will lead to something good.

PR #403, opened 2024-11-15
Upload test results to CodeCov

aik099:
Got an e-mail from CodeCov about a new `Test Analytics` feature (see https://about.codecov.io/blog/find-failing-and-flaky-tests-with-codecov-test-analytics/ ) and decided to try it out.

Doesn't work as expected (test results are not getting uploaded even with a CodeCov token). I've reported it to the CodeCov: https://github.com/codecov/test-results-action/issues/98 .


aik099:
I'm going to merge this because I think CodeCov doesn't like, that the updated workflow (where `CODECOV_TOKEN` is given to the new action) is absent in the `master` branch.

PR #406, opened 2025-01-11
Use "docker-composer.yml" file during the build

aik099:
Ported some changes from `webdriver-classic-driver`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,7 +39,7 @@ jobs:
 
   tests:
     name: "Tests (PHP ${{ matrix.php }}, Selenium ${{ matrix.selenium_version }})${{ matrix.with_coverage == true && ' with coverage' || ''}}"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         selenium_version: [ '2.53.1' ]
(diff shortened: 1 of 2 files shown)


aik099:
@uuf6429 , do you have an idea why this doesn't work? When Selenium tries to load the http://host.docker.internal:8002/js_test.html page from the Docker container it gets the `Unable to connect` page back, which is generated either by Firefox or Selenium Server (see in log: https://github.com/minkphp/MinkSelenium2Driver/actions/runs/12723843225/job/35469480643?pr=406 ).

uuf6429:
@aik099 my first guess would be something related to config, since the xml files are also a bit different:
- https://github.com/minkphp/MinkSelenium2Driver/blob/master/phpunit.xml.dist
- https://github.com/minkphp/webdriver-classic-driver/blob/main/phpunit.xml.dist

How I would troubleshoot it:
1. I would check the "PHP builtin server" (that mink launches) log - not sure if we store that somewhere..
2. since you set up the workflow for remote access, you can just `docker exec -ti` to get into the selenium container and ping that url you mentioned - if it works then probably it's some configuration problem

aik099:
> @aik099 my first guess would be something related to config, since the xml files are also a bit different:
> 
> * https://github.com/minkphp/MinkSelenium2Driver/blob/master/phpunit.xml.dist
> * https://github.com/minkphp/webdriver-classic-driver/blob/main/phpunit.xml.dist
> 
> How I would troubleshoot it:
> 
> 1. I would check the "PHP builtin server" (that mink launches) log - not sure if we store that somewhere..
> 2. since you set up the workflow for remote access, you can just `docker exec -ti` to get into the selenium container and ping that url you mentioned - if it works then probably it's some configuration problem

Thank you. The `docker exec -ti d026a18ffd9d bash` command revealed, that:

* Web Server was started on a `localhost:8002`
* docker container accessed Web Server via `172.17.0.1:8002`

That fixed Selenium 2 image usage, however, the Selenium 3 images keep failing with a different error.

aik099:
The Selenium 3 build is also fixed now.

Turned out, that the Selenium 3 container took a bit longer to start (then Selenium 2) and `while ! nc -z localhost 4444 </dev/null; do echo Waiting for remote driver to start...; sleep 1; done` command wasn't accounting for that (but I thought it should). The `curl --retry 5 --retry-all-errors --retry-delay 1 --max-time 10 --head -X GET http://localhost:4444/wd/hub/status` command, however, worked correctly. Mystery but it works now anyway.

@uuf6429 , Ready for review.

.github/workflows/tests.yml
```diff
@@ -74,24 +74,27 @@ jobs:
           composer update --no-interaction --prefer-dist
 
       - name: Setup Mink test server
+        env:
+          MINK_HOST: 0.0.0.0:8002
```
uuf6429, line 78: is this still needed then?

uuf6429 (approved):
One small comment - looks good otherwise.

.github/workflows/tests.yml
aik099: Yes, because otherwise Web Server would start on `localhost` but be accessed from within Docker container on `172.17.0.1`.

The similar approach is taken (replacement of `localhost` with `0.0.0.0`) in https://github.com/minkphp/webdriver-classic-driver/blob/main/tests/bootstrap.php#L13 .

uuf6429: AAah, I see.. this is listening host, not the connect-to host. All good then.

PR #407, opened 2025-01-11
Inform developers about non-working "rightClick" on Selenium Server 3.x

### LINKED ISSUE #381: The "EventsTest::testRightClick" test fails on Selenium 3.x
When the `\Behat\Mink\Tests\Driver\Js\EventsTest::testRightClick` test is executed on Selenium 3.x it fails with this message:

```
Failed asserting that two strings are equal.
Expected :'right clicked'
Actual   :'single clicked'
```

The driver code seems to be written correctly (I've double-checked with the Facebook's driver version: https://github.com/php-webdriver/php-webdriver/blob/5d8e66ff849b5614015d35e99d759252e371ce26/lib/Remote/RemoteMouse.php#L100 ).

Not sure why the test fails.

More research is done in https://github.com/minkphp/MinkSelenium2Driver/pull/354#issuecomment-1288121270 .

The fix was done in https://github.com/minkphp/MinkSelenium2Driver/pull/372 , but it's using Actions , which isn't implemented in the WebDriver.

aik099:
Closes #381

### DIFF
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -350,24 +350,29 @@ class Selenium2Driver extends CoreDriver
     {
         try {
             $status = $this->webDriver->status();
-            $majorSeleniumServerVersion = (int)explode('.', $status['build']['version'])[0];
+            $seleniumVersion = $status['build']['version'] ?? $status['nodes'][0]['version'] ?? 'unknown';
+            $seleniumMajorVersion = (int) explode('.', $seleniumVersion)[0];
+        } catch (\Throwable $ex) {
+            throw new DriverException("Selenium Server version could not be detected: {$ex->getMessage()}", 0, $ex);
+        }
 
-            if ($majorSeleniumServerVersion > 3) {
-                throw new \Exception(<<<TEXT
-The Selenium Server {$status['build']['version']} is not supported.
+        if ($seleniumMajorVersion > 3) {
+            throw new DriverException(<<<TEXT
+This driver requires Selenium version 3 or lower, but version {$seleniumVersion} was found.
 
 Please use the "mink/webdriver-classic-driver" Mink driver or switch to Selenium Server 2.x/3.x.
 TEXT
-                );
-            }
+            );
+        }
 
-            $this->isW3C = $majorSeleniumServerVersion === 3;
+        try {
+            $this->isW3C = $seleniumMajorVersion === 3;
             $this->wdSession = $this->webDriver->session($this->browserName, $this->desiredCapabilities);
 
             $this->applyTimeouts();
             $this->initialWindowHandle = $this->getWebDriverSession()->window_handle();
         } catch (\Exception $e) {
-            throw new DriverException('Could not open connection: '.$e->getMessage(), 0, $e);
+            throw new DriverException('Could not open connection: ' . $e->getMessage(), 0, $e);
         }
 
         $this->started = true;
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -69,6 +69,11 @@ class Selenium2Driver extends CoreDriver
     private $isW3C = false;
 
     /**
+     * @var bool
+     */
+    private $isRightClickSupported = false;
+
+    /**
      * The timeout configuration
      * @var array{script?: int, implicit?: int, page?: int}
      */
@@ -352,7 +357,11 @@ class Selenium2Driver extends CoreDriver
             $this->wdSession = $this->webDriver->session($this->browserName, $this->desiredCapabilities);
 
             $status = $this->webDriver->status();
-            $this->isW3C = version_compare($status['build']['version'], '3.0.0', '>=');
+            list($majorSeleniumServerVersion) = explode('.', $status['build']['version']);
+            $this->isW3C = (int)$majorSeleniumServerVersion >= 3;
+
+            // See: https://github.com/SeleniumHQ/selenium/commit/085ceed1f55fbaaa1d419b19c73264415c394905.
+            $this->isRightClickSupported = (int)$majorSeleniumServerVersion !== 3;
 
             $this->applyTimeouts();
             $this->initialWindowHandle = $this->getWebDriverSession()->window_handle();
@@ -936,6 +945,15 @@ JS;
 
     public function rightClick(string $xpath)
     {
+        if (!$this->isRightClickSupported) {
+            throw new DriverException(<<<TEXT
+The right-clicking functionality, via JsonWireProtocol, is not available on the Selenium Server 3.x.
+
+Please use "mink/webdriver-classic-driver" Mink driver or downgrade to Selenium Server 2.x.
+TEXT
+            );
+        }
+
         $this->mouseOver($xpath);
         $this->getWebDriverSession()->click(array('button' => 2));
     }
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -357,11 +357,9 @@ class Selenium2Driver extends CoreDriver
             $this->wdSession = $this->webDriver->session($this->browserName, $this->desiredCapabilities);
 
             $status = $this->webDriver->status();
-            list($majorSeleniumServerVersion) = explode('.', $status['build']['version']);
-            $this->isW3C = (int)$majorSeleniumServerVersion >= 3;
-
-            // See: https://github.com/SeleniumHQ/selenium/commit/085ceed1f55fbaaa1d419b19c73264415c394905.
-            $this->isRightClickSupported = (int)$majorSeleniumServerVersion !== 3;
+            $majorSeleniumServerVersion = (int)explode('.', $status['build']['version'])[0];
+            $this->isW3C = $majorSeleniumServerVersion >= 3;
+            $this->isRightClickSupported = $majorSeleniumServerVersion !== 3;
 
             $this->applyTimeouts();
             $this->initialWindowHandle = $this->getWebDriverSession()->window_handle();
@@ -946,10 +944,11 @@ JS;
     public function rightClick(string $xpath)
     {
         if (!$this->isRightClickSupported) {
+            // See: https://github.com/SeleniumHQ/selenium/commit/085ceed1f55fbaaa1d419b19c73264415c394905.
             throw new DriverException(<<<TEXT
-The right-clicking functionality, via JsonWireProtocol, is not available on the Selenium Server 3.x.
+Right-clicking via JsonWireProtocol is not possible on Selenium Server 3.
 
-Please use "mink/webdriver-classic-driver" Mink driver or downgrade to Selenium Server 2.x.
+Please use the "mink/webdriver-classic-driver" Mink driver or switch to Selenium Server 2 or 4+.
 TEXT
             );
         }
--- a/src/Selenium2Driver.php
+++ b/src/Selenium2Driver.php
@@ -69,11 +69,6 @@ class Selenium2Driver extends CoreDriver
     private $isW3C = false;
 
     /**
-     * @var bool
-     */
-    private $isRightClickSupported = false;
-
-    /**
      * The timeout configuration
      * @var array{script?: int, implicit?: int, page?: int}
      */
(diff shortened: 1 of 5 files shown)


src/Selenium2Driver.php
```diff
@@ -936,6 +945,15 @@ public function doubleClick(string $xpath)
 
     public function rightClick(string $xpath)
     {
+        if (!$this->isRightClickSupported) {
+            throw new DriverException(<<<TEXT
+The right-clicking functionality, via JsonWireProtocol, is not available on the Selenium Server 3.x.
```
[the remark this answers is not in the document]
aik099: Done.

I've used a link to commit instead of an issue because the commit clearly shows Selenium Server versions where the fix is available. In fact, I've already used that link above the `$this->isRightClickSupported` property assignment line, but I have now moved it here.

src/Selenium2Driver.php
```diff
@@ -352,7 +357,11 @@ public function start()
             $this->wdSession = $this->webDriver->session($this->browserName, $this->desiredCapabilities);
 
             $status = $this->webDriver->status();
-            $this->isW3C = version_compare($status['build']['version'], '3.0.0', '>=');
+            list($majorSeleniumServerVersion) = explode('.', $status['build']['version']);
```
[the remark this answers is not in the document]
aik099: Done.

src/Selenium2Driver.php
```diff
@@ -352,7 +357,9 @@ public function start()
             $this->wdSession = $this->webDriver->session($this->browserName, $this->desiredCapabilities);
 
             $status = $this->webDriver->status();
-            $this->isW3C = version_compare($status['build']['version'], '3.0.0', '>=');
+            $majorSeleniumServerVersion = (int)explode('.', $status['build']['version'])[0];
+            $this->isW3C = $majorSeleniumServerVersion >= 3;
```
uuf6429, line 361: I see that we already have a case where we talk to selenium with W3C... we could solve the right-click problem in the same way: when w3c is enabled, we use W3C actions.

We could do that by somehow retrieving the http client the instaclick library is using (it seems they have some sort of DIC, so maybe not too complicated).

Give me a moment and I'll check if that's possible.

uuf6429: > - right click works in Selenium Server 2 and 4+

That is because selenium 4 correctly translates old/JWP right-click call to W3C actions.
If we make an action call instead, it will fix the problem for selenium 3 and selenium 4 will be functionally the same (at least from browser perspective).

> Because of this, we can't really say, that right-click doesn't work Selenium Server 3+ as you're suggesting.

I've a feeling you're replying to the wrong comment, or I misunderstood. 😆

aik099: > ... I see that we already have a case where we talk to selenium with W3C...

My apologies. The `$this->isW3C` code made you think, that we support W3C commands, but we don't.

As you can see we're just checking if the Selenium Server version supports W3C. This was needed to supply correct timeout names (`pageLoad` instead of `page load` based on the Selenium Server version).

As I've written somewhere else the `instaclick/web-driver` doesn't have the public methods for calling any non-implemented APIs as `php-webdriver/webdriver` does.

Since we agreed to not support right-clicking on Selenium Server 3 specifically (and this PR does just that) I recommend not spending any more time here. Especially since the `webdriver-classic-driver` is available with both right-clicking and Selenium Server 4 support.

If you still, for example, to practice, want to figure out how to use internal classes of the `instaclick/web-driver` to invoke the right-click action, then you're welcome.

uuf6429: It's not possible.

The same actions request by the classic driver will cause an UnknownCommandError response with this driver - I was stumped, considering that as said it works with the other driver.

The reason turned out to be simple:
- first of all, we need to keep in mind that the protocol (W3C vs JWP/OSS) applies in different places:
  1. it applies between PHP (our drivers) and Selenium
  2. it also applies between Selenium and the web browser
  3. additionally, since Selenium 3, selenium translates from JWP to W3C in case of point ii (and when the browser is also communicating with W3C).
- when we request a new session from selenium, it determines the type of protocol for both the driver side and for the web browser.
- in this particular case and with this driver's desiredCapabilities, selenium thinks the driver wants to communicate with JWP, therefore W3C API calls (between our driver and selenium) are disabled - causing the UnkownCommandError mentioned above
- in case of the classic driver, the JWP API calls are disabled and the W3C ones enabled (opposite situation)

uuf6429: This means that we do not have any other options - if the session is started in JWP, there's no way to switch to W3C as far as I can see.

uuf6429 (approved):

src/Selenium2Driver.php
uuf6429: > My apologies. The $this->isW3C code made you think, that we support W3C commands, but we don't.

Yeah, I was going to suggest maybe renaming that property, something like `useW3CStyleTimeout`, or since this whole thing is the situation for the right-click check, maybe just make one private property (for both cases) called `majorSeleniumServerVersion` or even just `isSelenium3`. It would make the intention clearer and less guesswork.

aik099: * The `$this->isW3C` checks that Selenum Server Version >= 3
* The `$this->isRightClickSupported` checks that Selenum Server Version != 3

I can't really merge them, because they check different things :(

Maybe they can be merged (not sure) IF this driver won't work with Selenium Server 4+. Will it work with Selenium Server 4+?

uuf6429: > Will it work with Selenium Server 4+?

Actually no, I happened to try testing with selenium 4, but the driver would not even connect.

So those two do kinda align because:
- `>=4` driver doesn't even connect (not even sure if status endpoint works*)
- `>=3` right click will not work
- `3` timeout requires(I guess?) the W3C format?

*if it does, I believe we should throw an early exception that Selenium4 is definitely not supported.

aik099: > `>=4` driver doesn't even connect (not even sure if status endpoint works*)

The `status` works. I've checked today.

It might even create a session if you wrap all the desired capabilities in the `firstMatch` element, but that's super non-obvious and would break browser-specific capabilities. I'll update the error message to say `Selenium Server 3` instead of `Selenium Server 3+`.

> *if it does, I believe we should throw an early exception saying that Selenium4 is definitely not supported.

Good catch.

The earliest things would be upon an attempt to create a new session. In fact, I can do a `status` API call even for Selenium 4 and it passes. I can use that to decide if an exception should be thrown.

aik099: I've made a commit with the proposed changes.

Please review.

P.S.
There are no tests currently to confirm, that:

1. session can't be created on Selenium Server > 3.x
2. doing a right click on Selenium Server 3.x would throw an exception

Should I add them?

Even if we add them, then how do I test on Selenium 4 so that only single test is executed?

uuf6429:
> 1. session can't be created on Selenium Server > 3.x

Since you add a check that throws an exception for that, I would recommend adding a test - it should be easy to do, please it would prove that the `status` endpoint works.

> 3. doing a right click on Selenium Server 3.x would throw an exception

Yeah, I think it makes sense to test that too. Should be easy as well, probably you wouldn't even need to visit a page, even `about:blank` could work.

> Even if we add them, then how do I test on Selenium 4 so that only single test is executed?

In the driver test config, I would move `$majorSeleniumServerVersion = ...` to the top (it's a relatively fast operation, but you could also decide to cache it by setting it from constructor).

Then in `skipMessage()` something like:
```php
if($this->majorSeleniumServerVersion >= 4 && [$testCase, $test] !== [CustomWebDriverTest::class, 'testThatSeleniumMustBeLessThan4']) {
    return 'Selenium must be less than version 4';
}
```
on the whole should be pretty standard...unless I missed something.

src/Selenium2Driver.php
uuf6429: > 1. session can't be created on Selenium Server > 3.x

Since you add a check that throws an exception for that, I would recommend adding a test - it should be easy to do, please it would prove that the `status` endpoint works.

> 2. doing a right click on Selenium Server 3.x would throw an exception

Yeah, I think it makes sense to test that too. Should be easy as well, probably you wouldn't even need to visit a page, even `about:blank` could work.

> Even if we add them, then how do I test on Selenium 4 so that only single test is executed?

In the driver test config, I would move `$majorSeleniumServerVersion = ...` to the top (it's a relatively fast operation, but you could also decide to cache it by setting it from constructor).

Then in `skipMessage()` something like:
```php
// change the class name and method name to whatever you do to test point 2 above
if ($this->majorSeleniumServerVersion >= 4 && [$testCase, $test] !== [CustomWebDriverTest::class, 'testThatSeleniumMustBeLessThan4']) {
    return 'Selenium must be less than version 4';
}
```
on the whole should be pretty standard...unless I missed something.

uuf6429:
@aik099 I have some fixes built on top of your PR changes. How can I send the changes to you?

aik099:
> @aik099 I have some fixes built on top of your PR changes. ~How can I send the changes to you?~
> 
> Never mind, I made a patch... here you go: [tests-and-fixes.patch](https://github.com/user-attachments/files/18401392/tests-and-fixes.patch)
> 
> * `SeleniumSupportTest.php` - the 2 (new) tests mentioned earlier
> * `Selenium2Driver.php` - mainly, fixes for `start()` logic in particular:
>   
>   * throw `DriverException` instead of `\Exception`
>   * do not catch `DriverException` immediately while trying to start - it's pointless
>   * handle `status()` response structure in case of Selenium 4
> * `Selenium2Config.php` - many improvements:
>   
>   * refactored $testCase+$test usage
>   * improved selenium version handling + added utility methods
>   * utility methods are public, to be used in the test

@uuf6429 , I've tried to apply provided Patch in the PhpStorm and it had huge conflicts in several files. Are you sure, that you've created it based on last commit in this PR?

Anyway I've applied it, did some minor adjustments, pushed a few commits based on that.

Please see I haven't missed anything.

src/Selenium2Driver.php
```diff
@@ -349,15 +349,30 @@ private function executeJsOnElement(Element $element, string $script, bool $sync
...
     {
         try {
-            $this->wdSession = $this->webDriver->session($this->browserName, $this->desiredCapabilities);
-
             $status = $this->webDriver->status();
-            $this->isW3C = version_compare($status['build']['version'], '3.0.0', '>=');
+            $seleniumVersion = $status['build']['version'] ?? $status['nodes'][0]['version'] ?? 'unknown';
+            $seleniumMajorVersion = (int) explode('.', $seleniumVersion)[0];
+        } catch (\Throwable $ex) {
+            throw new DriverException("Selenium Server version could not be detected: {$ex->getMessage()}", 0, $ex);
```
aik099, line 356: @uuf6429 , how do we test this?

uuf6429: Sorry, I forgot to test that one :(

Easiest way I can think of is to mock `$this->webDriver->status()` response.

uuf6429: @aik099 here you go, this works for me:
```php
    public function testThatErrorsWhileRetrievingVersionAreCaught(): void
    {
        $mockWebDriver = $this->createPartialMock(WebDriver::class, ['status']);
        $mockWebDriver
            ->method('status')
            ->willThrowException(new \RuntimeException('some internal error'));

        $driver = $this->createDriver();
        if (!$driver instanceof Selenium2Driver) {
            $this->markTestSkipped(sprintf('This test applies to %s only, found %s instead.', Selenium2Driver::class, get_class($driver)));
        }
        $driver->setWebDriver($mockWebDriver);

        $this->expectException(DriverException::class);
        $this->expectExceptionMessage('Selenium Server version could not be detected: some internal error');

        $driver->start();
    }
```

I would put it in `WebDriverTest``.

uuf6429: @aik099 Hey! Hope you're doing well.

Just a small reminder, any news on this PR (in particular, the last test mentioned above)?
I believe that's the only remaining thing in getting this PR merged, right?

aik099: I planned to write that test this week but lacked the time. I appreciate you did it for me.

I've made some adjustments, though:
* test rename
* removed usage of `createPartialMock` method, because it was adding a warning (attempt to use a deprecated method) and wasn't clearly expressing a fact, that `status` method doesn't really exist in WebDriver class
* used `new Selenium2Driver` construct to avoid `$this->createDriver() + checking if that's a Selenium2Driver class being used in the Selenium2Driver only test suite` check, which I thought was added just to make PhpStan happier :)
.

Pushed now.

aik099:
@uuf6429 , ready for review.

uuf6429 (approved):

PR #409, opened 2025-01-13
Also test on Chrome for better coverage

aik099:


aik099:
After looking at https://github.com/minkphp/MinkSelenium2Driver/actions/runs/12745224912/job/35518756018 (Selenium 3 + Chrome) I now realize why there was an issue reported some time ago about fixing key-related event emulation. But I can't seem to find it.

PR #411, opened 2025-01-20
Replaced short array syntax usages with traditional array syntax

aik099:
The short array syntax was accidentally introduced in the #407.

If there is a need, then we can mass-convert driver code to use short array syntax (e.g. using PHP_CodeSniffer rule or PhpStorm inspection).

### DIFF
--- a/tests/Selenium2Config.php
+++ b/tests/Selenium2Config.php
@@ -51,9 +51,9 @@ class Selenium2Config extends AbstractConfig
 
     public function skipMessage($testCase, $test): ?string
     {
-        $testCallback = [$testCase, $test];
+        $testCallback = array($testCase, $test);
 
-        if ([Html5Test::class, 'testHtml5Types'] === $testCallback) {
+        if (array(Html5Test::class, 'testHtml5Types') === $testCallback) {
             return <<<TEXT
 WebDriver does not support setting value in color inputs.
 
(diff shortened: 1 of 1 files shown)


uuf6429 (approved):
I'm not a fan of this change since:
- in my eyes, it's a downgrade
- it still isn't the proper way of telling devs what kind of coding style is allowed
- it would have been automated and more future-proof if done through PHP-CS-Fixer (or similar)

In practices, since it's harmless, I'm fine with approving it.

aik099:
> * in my eyes, it's a downgrade

Syntax is syntax. Use whichever that works but be consistent throughout the codebase. This wasn't the case, hence this PR.

Anyway this PR is a transient step (as I've probably mentioned elsewhere) to keep syntax consistent. In another PR you can perform system-wide (driver code & driver tests) syntax change to short array declaration (you can use tools shown in the PR description).

> * it still isn't the proper way of telling devs what kind of coding style is allowed
> * it would have been automated and more future-proof if done through PHP-CS-Fixer (or similar)

Completely agree. I'm open to suggestions. Please create an issue where we can discuss our options. We can later apply same concepts to the `mink/webdriver-classic-driver` as well.

## In an ideal world:

* use a consistent code style across the codebase
* stick to a coding standard (the PSR2 is used currently)
* use a more strict coding standard
* fail the build when PR code doesn't follow the standard

## What we have now:

* the PSR2 standard doesn't care about array declarations
* coding style isn't automatically checked for PRs

.

PR #414, opened 2025-11-22
Add a CI job running on PHP 8.5

stof:


aik099:
I guess there will be another PR to the test suite that will skip usage of the `imagedestroy` and `curl_close` functions on PHP > 8.0. Without this build for PHP 8.5 now fails.

stof:
The deprecation of `imagedestroy` has been fixed in https://github.com/minkphp/driver-testsuite/pull/107.
The deprecation of `curl_close` will be fixed once https://github.com/instaclick/php-webdriver/pull/141 is merged and released.

